PR #45, opened 2018-07-05
1. Enable loading both nwb ver 102 and 105. 2. Refactored the qc modu…

sgratiy:
…le by splitting feature calculations from comparing them to the criteria. 3.Eliminated the need for the MiesDataSet class. 4. Grouped functions for returning stim intervals into a single module.
Sorry for such a large commit.


dyf:
woops, didn't mean to close the PR just now.  sorry.

Also @sgratiy, make sure you run py.test to see if the existing tests pass since we don't have CI set up yet.

PR #48, opened 2018-07-13
47/ci pytests

sgratiy:


tests/test_pipeline_output.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+    """
+    specimens = np.array([500844783, 509604672, 513876493])
+
+    validation = []
+
+    for specimen in specimens:
+        print("testing specimen %s" % specimen)
+        benchmark_dir = "/allen/aibs/informatics/module_test_data/ipfx/benchmark_pipeline_output/%s" % specimen
```
dyf: Don't hardcode these paths into your script.  These need to be configurable.  Do this by passing in a configuration file in an environment variable and putting this at the top of this file:

```python
from pkg_resources import resource_filename
import resource_package
if 'TEST_SPECIMEN_FILE' in os.environ:
    TEST_SPECIMEN_FILE = os.environ['TEST_SPECIMEN_FILE']
else:
    TEST_SPECIMEN_FILE = resource_filename(__name__, 'test_specimens.txt')

with open(TEST_SPECIMEN_FILE,'rb') as f:
    TEST_SPECIMENS = list(f)
```

tests/test_pipeline_output.py
```diff
@@ -0,0 +1,71 @@
...
+
+def test_pipeline_output():
+    """
+    Validates pipeline output for a list of specimens.
+    Runs pipeline, saves data to disk and compares pipeline output to that from the benchmark runs.
+    If pipeline output has changed (was not validated), raises assertion error and lists specimens failing validation.
+
+
+    """
+    specimens = np.array([500844783, 509604672, 513876493])
```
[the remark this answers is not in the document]
dyf: If you do the `TEST_SPECIMEN_FILE` thing I said below, this turns into:
```python
@pytest.mark.parametrize('specimen_id', TEST_SPECIMENS)
def test_pipeline_output(specimen_id):
   # don't loop in here
```

tests/test_pipeline_output.py
dyf: Also, your `TEST_SPECIMENS` should not be a list of IDs that you append to a directory -- it should be fully-qualified paths to `pipeline_input.json`.

tests/test_pipeline_output.py
```diff
@@ -0,0 +1,71 @@
...
+
+        test_pipeline_output_json = os.path.join(test_dir,"pipeline_output.json")
+        ju.write(test_pipeline_output_json, test_pipeline_output)
+
+        benchmark_pipeline_output_json = os.path.join(benchmark_dir, "pipeline_output.json")
+
+        val = check_json_files_are_equal(benchmark_pipeline_output_json, test_pipeline_output_json)
+        validation.append(val)
+
+    specimens_failing_validation = specimens[~np.array(validation)]
```
dyf: if you `parametrize` the test, you don't need to keep track of this list (yay!)

tests/test_pipeline_output.py
```diff
@@ -0,0 +1,71 @@
+import pytest
+import numpy as np
+import os
+import allensdk.core.json_utilities as ju
+
+
+import sys
+sys.path.append("/local1/pyprojects/allensdk.ipfx/bin") # there is got to be a better way
```
dyf: move the `bin` directory to `allensdk.ipfx/allensdk/ipfx/bin` and add an `__init__.py`.  Then you can do this:

```python
from allensdk.ipfx.bin.run_pipeline import run_pipeline
```

tests/test_pipeline_output.py
```diff
@@ -0,0 +1,71 @@
...
+
+    """
+    specimens = np.array([500844783, 509604672, 513876493])
+
+    validation = []
+
+    for specimen in specimens:
+        print("testing specimen %s" % specimen)
+        benchmark_dir = "/allen/aibs/informatics/module_test_data/ipfx/benchmark_pipeline_output/%s" % specimen
+        test_dir = "/allen/aibs/informatics/module_test_data/ipfx/test_pipeline_output/%s" % specimen
```
dyf: Use `tmpdir_factory` for this.  You can see how @NileGraddis did it here:
https://github.com/AllenInstitute/AllenSDK/blob/b2d0bac25a9e2b6a587be9d7902a25d0e6bfe936/allensdk/test/test_inline_examples.py#L20

the `tmpdir_factory` object allows you to create a temporary directory and by default these will go into `tmp` that will get cleaned up.  However, you can override the default temp directory from the command line like this:

`pytest --basetemp=some_dir`

https://docs.pytest.org/en/latest/tmpdir.html

So bamboo can keep the files or not.

dyf: I expect this will look something like this:
```python
@pytest.mark.parametrize('specimen_id,test_pipeline_input', zip(TEST_PIPELINE_INPUTS, TEST_SPECIMEN_IDS))
def test_pipeline_output(specimen_id, test_pipeline_input, tmpdir_factory): # tmpdir_factory should just exist
   #...
   test_pipeline_output = tmpdir_factory.mktemp("data").join('%d_pipeline_output.json' % specimen_id)
   # ...
```

tests/test_pipeline_output.py
```diff
@@ -0,0 +1,67 @@
...
+    -------
+
+    """
+
+    print (specimen,benchmark_pipeline_input_json,benchmark_pipeline_output_json)
+
+    bpi = ju.read(benchmark_pipeline_input_json)
+    test_dir = str(tmpdir_factory.mktemp("%s" % specimen))
+
+    tpi = dict(bpi)
```
dyf: I think you're trying to make a copy here.  1) This will only do a shallow copy -- you probably want `copy.deepcopy`.  2) Do you need to make a copy?  `bpi` isn't used anywhere else.

Issue #56, opened 2018-07-31
Spurious spike as a result of trace cut off


sgratiy:
Cut off response traces are detected as spikes as well as issuing a warning stating that voltage did not return to resting value. Need to exclude zeros from the traces passed the end of the recording.

Issue #50, opened 2018-07-17
Membrane time constant fit does not meet tolerance


sgratiy:
Commonly the RMSE for the fit is much greater than the RMSE preventing calculation of the membrane time constant.

gouwens:
Could you clarify this a bit more? The check is there so that bad fits won't be included in the average time constant for the cell. Is the issue (1) that all the subthreshold sweeps are failing the RSME check, or (2) that it seems like a lot of individual sweeps fail this check?

Option (1) could be a problem if we are too strict for too many cells, but (2) is only a problem (I think) if there are good fits that are still getting rejected.

PR #57, opened 2018-07-31
Clipped spikes

sgratiy:
Resolve issue #56 (as well as #50). Now the code checks whether the sweep is truncated and for the unctruncated (i.e., completed sweeps) uses only recorded portion of the sweep (i.e., cut off trailing zeros).

### DIFF
--- a/allensdk/ipfx/data_set_features.py
+++ b/allensdk/ipfx/data_set_features.py
@@ -41,6 +41,7 @@ from . import ephys_extractor as efex
 from . import ephys_features as ft
 from . import ephys_data_set as eds
 from . import stimulus_protocol_analysis as spa
+import stim_features as st
 
 DEFAULT_DETECTION_PARAMETERS = { 'dv_cutoff': 20.0, 'thresh_frac': 0.05 }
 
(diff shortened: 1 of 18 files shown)


allensdk/ipfx/data_set_features.py
```diff
@@ -171,24 +173,29 @@ def extract_cell_features(data_set,
...
 
     # short squares
+    logging.info("Analyzing Short Square")
     if len(short_square_sweep_numbers) == 0:
         raise ft.FeatureError("no short square sweep numbers provided")
 
     ssq_sweeps = data_set.sweep_set(short_square_sweep_numbers)
-    ssq_start, ssq_dur, _, _, _ = get_stim_characteristics(ssq_sweeps.sweeps[0].i, ssq_sweeps.sweeps[0].t)
+
+    ssq_start, ssq_dur, _, _, _ = st.get_stim_characteristics(ssq_sweeps.sweeps[0].i, ssq_sweeps.sweeps[0].t)
```
dyf: I don't think you can assume that all short squares have the same start time anymore.  You need to align all of the timestamps so that they have the same estimation window.

sgratiy: I would prefer to deal with alignment on a separate PR.

dyf:
approved -- taking stimulus alignment into a separate PR

PR #59, opened 2018-08-14
Add conversion from abf to nwb

t-b:


t-b:
@dyf Something like the included readme?

dyf:
Thanks @t-b !

PR #64, opened 2018-08-27
Feature/add mcc settings getter

t-b:
This is the first version of the MCC settings gatherer. It loops over all connected amplifiers and stores all settings we currently store in MIES.

This is a very basic  version so feel free to ask for more features.


t-b:
@dyf Let's merge now (I don't have enough power to do that.).

PR #79, opened 2018-09-28
Update module schema

sgratiy:
1. Updated module schema and scripts generating module input json 
2. Moved figure plotting back into FX module 
3. when qc_figs_dir is None will not plot figures


ipfx/_schemas.py
```diff
@@ -4,9 +4,11 @@
 
 class SweepFeatures(DefaultSchema):
     stimulus_code = String(description="stimulus code", required=True)
+    stimulus_name = String(description="stimulus name", required=True)
```
dyf: You should be able to get this (`stimulus_name`) from the ontology instead of the schema, no?

sgratiy: Removed stimulus_name from the schema

PR #92, opened 2018-10-16
Add sweep info

t-b:
See ba098d8


dyf:
This is fine for now.  We'll work on migrating this information into an explicit sweep table when pynwb is ready.

PR #99, opened 2018-10-31
Fix checking for clipped spikes

sgratiy:
Moved clipped spikes checking into a function
Avoid running and failing py3 tests
substitutes deprecated .ix for .loc in pandas


dyf:
@nikahk the `v` `t` thing came from the original design years ago.

gouwens:
I'd argue that in this case the short names are preferable (cf. https://softwareengineering.stackexchange.com/questions/176582/is-there-an-excuse-for-short-variable-names)

sgratiy:
At least for me short names work fine here: they follow a standard physics notation that I can understand.

PR #112, opened 2018-11-20
FX queue fix

sgratiy:
1. Update FX queue input; 2. Refactor generator scripts for each queue to use the same input parameters


ipfx/bin/pipeline_from_nwb_file.py
```diff
@@ -1,41 +1,50 @@
...
-from ipfx._schemas import GeneratePipelineInputParameters
+import sys
 import os.path
 from run_pipeline import run_pipeline
 import generate_pipeline_input as gpi
 
+OUTPUT_DIR = "/local1/ephys/patchseq/tsts"
 
 INPUT_JSON = "pipeline_input.json"
 OUTPUT_JSON = "pipeline_output.json"
```
sgratiy: I can not get logging output when running from python. Is there a way to enable logging?

ipfx/bin/generate_pipeline_input.py
```diff
@@ -68,50 +68,63 @@ def get_input_h5_file_from_lims(specimen_id):
...
+    pipe_input['output_nwb_file'] = os.path.join(args["cell_dir"], "output.nwb")
+    pipe_input['qc_criteria'] = ju.read(qcp.DEFAULT_QC_CRITERIA_FILE)
+
+    return pipe_input
 
 
 def main():
 
+    """
+    Usage:
```
sgratiy: I wanted to be able to run either starting with specimen_id, or nwb_file and in the future perhaps could add ephys_roi_result. Is this the right way to do this?

ipfx/bin/generate_pipeline_input.py
```diff
@@ -68,50 +68,63 @@ def get_input_h5_file_from_lims(specimen_id):
     return h5_file_name
 
 
-def make_input_json(args, plot_figures=False):
```
sgratiy: I wonder if it is Ok to pass arguments this way when args here are actually keyword arguments

PR #113, opened 2018-11-22
mcc_get_settings.py: Add option to read the channel mapping from a JSON file

t-b:
The quoting required from passing in a valid JSON file on the command
line is just too difficult to get right.

Now the user can pass in a filename where a JSON encoded file is read.


t-b:
Feedback from users is that it works.

PR #116, opened 2018-11-28
Eliminate "truncated" property 

sgratiy:


ipfx/sweep_props.py
```diff
@@ -0,0 +1,51 @@
...
+        sn = sweep["sweep_number"]
+        if sn in sweep_states:
+            sweep["passed"] = sweep_states[sn]
+        else:
+            logging.warning("could not find QC state for sweep number %d", sn)
+
+
+def drop_incomplete_sweeps(sweep_features):
+
+    sweep_features[:] = [s for s in sweep_features if s["completed"]]
```
nikahk: In general, it's better to have functions that have a single purpose
The assign_sweep_states is doing two things. Assigning sweep_states andd sweep["passed]

ipfx/qc_features.py
```diff
@@ -287,82 +284,112 @@ def sweep_qc_features(data_set):
...
+
+    post_stim_response_duration = (response_end_ix - stimulus_end_ix) / hz
+    completed_expt_epoch = post_stim_response_duration > POSTSTIM_STABILITY_EPOCH
+
+    long_response = response_end_ix/hz>LONG_RESPONSE_DURATION
+
+    if completed_expt_epoch or long_response:
+        sweep_completion = True
+    else:
+        sweep_completion = False
```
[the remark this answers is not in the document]
njmei: It might be even better to just make it a ternary?

`sweep_completion = (True if completed_expt_epoch or long_response else False)`

ipfx/sweep_props.py
sgratiy: Done. thnx.

PR #117, opened 2018-11-30
Nwbreader fix

sgratiy:


dyf:
this looks like it might break @t-b .  adding him as a reviewer.

ipfx/nwb_reader.py
```diff
@@ -415,24 +410,43 @@ def get_stim_code(self, sweep_name):
...
-
-        # but in V2 this is an attribute
-        nwb_version = f["/"].attrs.get("nwb_version")
-        if nwb_version is not None and re.match("^2", nwb_version):
-            return {"major": 2, "full": nwb_version}
+        if "nwb_version" in f:         # In v1 this is a dataset
+            nwb_version = f["nwb_version"].value
+            if isinstance(nwb_version, np.ndarray):
+                nwb_version = nwb_version[0]
+            if nwb_version is not None and re.match("NWB-1", nwb_version):
```
t-b: Thanks for catching that the nwb version string start with `NWB-`. Can you make that `"^NWB-1"` so that it's clear that it has to be at the beginning?

ipfx/nwb_reader.py
```diff
@@ -415,24 +410,43 @@ def get_stim_code(self, sweep_name):
...
+        sweep_naming_convention = sweep_names[0].split('_')[0]
+
+        if sweep_naming_convention == "Sweep":
+            nwb_flavor = "Pipeline"
+
+        elif sweep_naming_convention == "data":
+            nwb_flavor = "Mies"
+
+        else:
+            raise ValueError("Unknown sweep naming convention: %s" %sweep_naming_convention)
```
t-b: flake8 complains here with `nwb_reader.py:445:69: E225 missing whitespace around operator`. Not sure if you follow these rules here or not.

ipfx/nwb_reader.py
```diff
@@ -415,24 +410,43 @@ def get_stim_code(self, sweep_name):
...
 
 
+def get_nwb_flavor(nwb_file):
+
+    with h5py.File(nwb_file, 'r') as f:
+        sweep_names = [e for e in f["acquisition/timeseries"].keys()]
+        sweep_naming_convention = sweep_names[0].split('_')[0]
+
+        if sweep_naming_convention == "Sweep":
+            nwb_flavor = "Pipeline"
```
t-b: You can immediately return here and in the below `elif` branch.

ipfx/nwb_reader.py
```diff
@@ -446,21 +460,15 @@ def create_nwb_reader(nwb_file):
     """
 
     nwb_version = get_nwb_version(nwb_file)
+    nwb_flavor = get_nwb_flavor(nwb_file)
```
t-b: You don't need to call `get_nwb_flavour` here, but only below in the `elif nwb_version["major"] == 1:` branch.

t-b:
@sgratiy @dyf I'm working on some basic NWBReader tests so that we can be sure that stuff does not break so easily.

t-b:
@sgratiy 
> stimlus_path is "stimulus/timeseries" for NwbPipelineReader (which reads data from the epochs group) and "stimulus/presentation" for NwbMiesReader. So, unfortunately, a single stimulus_path string would not work. What do you think?

In that case we can either feed in the `nwb_flavour` into NwbReader ctor or every derived ctor defines these strings on their own.

sgratiy:
@t-b 
I would prefer the latter option, i.e.,
"derived ctor defines these strings on their own"

t-b:
@sgratiy 

> "derived ctor defines these strings on their own"

Sounds good for me. Are you doing that right now or should I prepare a fix?

sgratiy:
I will take care of the strings today.

sgratiy:
On the second thought it seems not straight forward to define string paths because e.g. for acquisition:
in the NwbPipelineReader we get acquisition from the epochs/Experiment_#/response/timeseries
in NwbMiesReader it is in acquisition/timeseries/data_#
Thus the sweep number is in middle of the string for the NwbPipelineReader and at the end for the NwbMiesReader. So, If you have a solution for this, pls go ahead and implement it.

sgratiy:
@t-b 
I think I addressed all of your suggestions in the latest commit.

t-b:
@sgratiy Thanks. Looks good.

t-b:
@sgratiy @dyf I've pushed some more changes and tests for NwbReader and the derived classes. This now works with the v1 files I have access to and the v2 files (from ABF and DAT). If there are some v1 which don't yet work I'm happy to fix it before merging.

t-b:
If tests for the NWBReader are correct we have a coverage of 88% via

```
pip install pytest-cov
coverage run -x -m pytest test_nwb_reader.py
```

which is okay.

sgratiy:
@t-b,
We need to make sure it works for the Mies nwb files. Pls remind how could I best share these with you?

t-b:
@sgratiy I've used a MIES NWB file which was created with current master version. For sharing files with me see https://ai-mies.slack.com/archives/DB81GAXL5/p1542739139000500.

sgratiy:
I have created an ipfx_nwb folder in that dropbox account and dropped there 3 Mies nwb files.

sgratiy:
I am a bit confused about " MIES NWB file which was created with current master version". Mies nwb files are created by mies during recordings. Have you used some kind of converter?
I think if the code runs with the Mies files that I provided, then we should be safe to (rebase+merge).

t-b:
> " MIES NWB file which was created with current master version".

I acquired some data here with the latest master version of the MIES software package (I'm also developing the MIES software package). And then converted that to NWBv1. And then added a test based on that file.

ipfx/nwb_reader.py
```diff
@@ -318,24 +318,26 @@ def get_sweep_number(self, sweep_name):
 
     def get_stim_code(self, sweep_name):
 
-        stimulus_description = "aibs_stimulus_description"
+        names = ["aibs_stimulus_name", "aibs_stimulus_description"]
```
sgratiy: I think Pipeline nwbs only use "aibs_stimulus_name"

sgratiy:
Ok, that makes sense.

ipfx/nwb_reader.py
t-b: @sgratiy The file 595570553.nwb which I'm using from the celltypes database is from the pipeline and has `aibs_stimulus_name` see

```
h5dump 595570553.nwb | grep -A 3 stimulus_name
           DATASET "aibs_stimulus_name" {
              DATATYPE  H5T_STRING {
                 STRSIZE H5T_VARIABLE;
                 STRPAD H5T_STR_NULLTERM;
                 CSET H5T_CSET_UTF8;
                 CTYPE H5T_C_S1;
```

and https://github.com/AllenInstitute/ipfx/blob/fef94b981ed174fd7820b72abb353e3181937d72/tests/test_nwb_reader.py#L130.

PR #119, opened 2018-12-06
More robust time constant estimation

gouwens:
Improve estimation of time constants for cells by
1. Rejecting sweeps with low signal-to-noise
2. Using a consistent time window for the exponential fits

I also added more documentation to the `time_constant()` function and added a few related tests.


gouwens:
Okay, I think the issue for me was that all those test_pipeline_output.py tests (I think 5/5) were failing for me on my machine even before I made changes. I had thought that was due to ongoing nwb reader work, but maybe not.

When I tried just now to make an AibsDataSet for specimen 509604672, it didn't work (error somewhere in h5py), so I can't look at those traces right now. Is the test looking at taus on a sweep-by-sweep basis, or for the whole cell? If the former, it could make sense that you now get a None/NaN instead of a value, since part of the change is to not even try to fit sweeps that are too noisy to expect a reliable estimate.

What changes should I make?

sgratiy:
Indeed, regression tests were broken due to nwb_reader updates. But now regression tests are passing. Thus, if you update your fork, you should be able to run the tests to see the errors. 
One way to run:
bash pipeline_from_specimen_id.sh 509604672
and specify the OUTPUT_DIR inside the script.

Regression tests compare all values from the pipeline output json, thus include both sweep-by-sweep and whole cell values. The None happened for the sweep value. 

I would prefer if you could confirm that you expect the changes to taus where they occurred and that the new values are acceptable. Then, I could update the regression tests to treat new pipeline outputs as as new ground truths.

gouwens:
Great, thanks for the info. I’ll try to get those running on my end and confirm that the new values (including None’s) are appropriate.

gouwens:
I've checked all 5 regression test specimens with the new code and get appropriate numbers for all their time constants. Two of them have one or two sweeps that now return NaN values (because they have too much noise).

So I think now it's just a matter of updating the tests.

PR #120, opened 2018-12-06
More nwbreader tests

t-b:
@dyf @sgratiy I've added two more MIES files from the three which Sergey passed onto me. I did not include the third one as that was quite similiar to the existing ones.


sgratiy:
Two of the nwb_reader tests are failing. Could you please take a look?
../../tests/test_nwb_reader.py ..........FF

PR #122, opened 2018-12-08
Ontology fix:

sgratiy:


sgratiy:
LIMS generates stimuli mapping and making ontology in generate_se_input.py and generate_fx_input.py respectively.

PR #127, opened 2018-12-11
Add ephys dataset tests

t-b:
This restructures the EphysDataSet and AibsDataSet classes and adds tests for EphysDataSet. IOW this is the groundwork for the new EphysDataSet class which reads NWBv2 data.

### DIFF
--- a/ipfx/aibs_data_set.py
+++ b/ipfx/aibs_data_set.py
@@ -2,6 +2,7 @@ import pandas as pd
 import logging
 
 from .ephys_data_set import EphysDataSet
+
 import ipfx.lab_notebook_reader as lab_notebook_reader
 import ipfx.nwb_reader as nwb_reader
 
@@ -93,10 +94,7 @@ class AibsDataSet(EphysDataSet):
             sweep_record["stimulus_code"] = stim_code
 
             if self.ontology:
-                # make sure we can find all of our stimuli in the ontology
-                stim = self.ontology.find_one(stim_code, tag_type='code')
-                sweep_record["stimulus_name"] = stim.tags(tag_type='name')[0][-1]
-
+                sweep_record["stimulus_name"] = self.get_stimulus_name(stim_code)
 
             sweep_props.append(sweep_record)
 
--- a/ipfx/aibs_data_set.py
+++ b/ipfx/aibs_data_set.py
@@ -10,9 +10,11 @@ import ipfx.nwb_reader as nwb_reader
 class AibsDataSet(EphysDataSet):
     def __init__(self, sweep_info=None, nwb_file=None, h5_file=None, ontology=None, api_sweeps=True):
         super(AibsDataSet, self).__init__(ontology)
-        self.nwb_file = nwb_file
-        self.h5_file = h5_file
+
         self.nwb_data = nwb_reader.create_nwb_reader(nwb_file)
+
+        if nwb_file and h5_file:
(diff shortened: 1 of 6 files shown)


tests/helpers_for_tests.py
```diff
@@ -0,0 +1,38 @@
+import numbers
+
+import numpy as np
+import pytest
+from pytest import approx
+
+
+def compare_dicts(d_ref, d):
```
sgratiy: I am using dictdiffer library to compare dictionaries (see test_pipeline_output.py) for this. Is there a reason why you would use a custom function instead?

ipfx/aibs_data_set.py
```diff
@@ -70,101 +68,35 @@ def extract_sweep_meta_data(self):
...
 
             sweep_record["stimulus_code_ext"] = stim_code_ext
             sweep_record["stimulus_code"] = stim_code
 
-            if self.ontology:
-                # make sure we can find all of our stimuli in the ontology
-                stim = self.ontology.find_one(stim_code, tag_type='code')
-                sweep_record["stimulus_name"] = stim.tags(tag_type='name')[0][-1]
-
+            self.handle_ontology(sweep_record)
```
sgratiy: Here we set stimulus_name and "handle_ontology" function obscures the task. Perhaps something like this would be more explicit:
sweep_record["stimulus_name"] = get_stimulus_name(stimulus_code)

ipfx/ephys_data_set.py
```diff
@@ -79,21 +86,119 @@ def get_sweep_info_by_sweep_number(self, sweep_number):
...
+        Parameters
+        ----------
+        sweep_number
+
+        Returns
+        -------
+        Sweep object
+        """
+
+        sweep_data = self.get_sweep_data(sweep_number)
```
sgratiy: Nice move!

tests/helpers_for_tests.py
t-b: @sgratiy Thanks for the pointer. I was not aware of dictdiffer and my first tests are not promising unfortunately.

There is a bug in dictdiffer as it can not handle numpy datatypes properly as

```
In [58]: list(diff({"a": np.float32('nan')}, {"a" : float('nan')}))
Out[58]: [('change', 'a', (nan, nan))]

In [60]: are_different(np.float32('nan'),np.nan, 0)
Out[60]: True
```

shows. We get the np.float32('nan') value from the NWB file.

I'll file a bug report with them ASAP.

ipfx/aibs_data_set.py
t-b: Good point. Will change.

t-b: @sgratiy This will change the logic though as with your proposal `sweep_record["stimulus"name"]` will exist and hold None without ontology where as the entry did not exist before. I guess we can live with that change though or?

sgratiy: how about this?:
if self.ontology:
      sweep_record["stimulus_name"] = get_stimulus_name(stimulus_code)

tests/helpers_for_tests.py
t-b: I've documented in 7e770b2d (Tests/compare_dicts: Explain why we don't use dictdiffer here, 2018-12-17) why dictdiffer does not work.

ipfx/aibs_data_set.py
t-b: Done.

t-b:
@sgratiy Ready for next review round.

PR #135, opened 2018-12-14
Fx state

sgratiy:


sgratiy:
I still need to update the regression tests, thus please do not merge. I will update test output shortly.

t-b:
We should get the bamboo tests integrated so that the PRs are checked before we merge. https://blog.buildo.io/github-bamboo-ci-6f681b00d310 has description on how to do that, we will use that as well as soon as we move the MIES [1](https://github.com/AllenInstitute/MIES/issues/1) repo onto github as well.

sgratiy:
That is a good idea and I will try setting this up in the near future. Thanks for the info.

PR #136, opened 2018-12-17
run_x_to_nwb_conversion: Ignore feedback stimulus data by default for…

t-b:
… ABF files

Not all data from ABF files is real data, some is also the fedback
stimulus data.

Ignore that by default when converting.

The distinction can be made due to the following properties:

> The best way to distinguish are the names. The real data channels are
> always "IN0", "IN1", "IN2" or "IN3". The feedback channels are "IN4"
> (for IN0), "IN5" (for IN1), "IN6" (for IN2) and "IN7" (for IN3). There
> is one set up where the real data channel is called "IN0" and the
> feedback data channel is called "RawOutput".

Source: Private communication from Natalia Goriounova (Mansvelder lab).


ipfx/x_to_nwb/ABFConverter.py
```diff
@@ -423,6 +426,16 @@ def _createAcquiredSeries(self, electrodes):
             for sweep in range(abf.sweepCount):
                 cycle_id = createCycleID([file_index, sweep], total=self.totalSeriesCount)
                 for channel in range(abf.channelCount):
+                    adcName = abf.adcNames[channel]
+
+                    if not self.outputFeedbackChannel:
+                        if adcName in ("IN 0", "IN 1", "IN 2", "IN 3"):
```
dyf: maybe move the definition of this tuple out of the inner loop and keep it as a constant instance variable?

t-b: Done!

PR #150, opened 2019-01-17
Fix bugs in stim start

sgratiy:
Eliminate redundant find_stim_start function and use consistently a single function for finding stim indexes.
Split epochs detection into a separate module

### DIFF
--- a/ipfx/epochs.py
+++ b/ipfx/epochs.py
@@ -1,0 +1,132 @@
+import numpy as np
+import time_series_utils as tsu
+
+
+# global constants
+#TODO: read them from the config file
+
+NOISE_EPOCH = 0.0015
+PRESTIM_STABILITY_EPOCH = 0.5
+POSTSTIM_STABILITY_EPOCH = 0.5
+LONG_RESPONSE_DURATION = 5  # this will count long ramps as completed
+
+
+def get_last_vm_epoch(idx1, hz):
+    """
+    Get epoch lasting LAST_STABILITY_EPOCH before the end of recording
+
+    Parameters
+    ----------
+    idx1
+    hz
+
+    Returns
+    -------
+
+    """
+    return idx1-int(POSTSTIM_STABILITY_EPOCH * hz), idx1
+
+
+def get_first_vm_noise_epoch(idx, hz):
+
+    return idx, idx + int(NOISE_EPOCH * hz)
+
+
+def get_last_vm_noise_epoch(idx1, hz):
+
+    return idx1-int(NOISE_EPOCH * hz), idx1
+
+
+def get_stability_vm_epoch(stim_start, hz):
+    num_steps = int(PRESTIM_STABILITY_EPOCH * hz)
+    if num_steps > stim_start-1:
+        num_steps = stim_start-1
+    elif num_steps <= 0:
+        return 0, 0
+    assert num_steps > 0, "Number of steps should be a positive integer"
+
+    return stim_start-1-num_steps, stim_start-1
+
+
+def get_recording_end_idx(v):
+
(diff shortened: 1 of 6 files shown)


ipfx/qc_features.py
```diff
@@ -371,23 +371,33 @@ def current_clamp_sweep_qc_features(sweep_data,sweep_info,ontology):
...
+    stim_start_time, stim_duration, stim_amplitude, stim_start_idx, stim_end_idx = stf.get_stim_characteristics(i,t)
+    expt_start_ix, expt_end_ix = stf.get_experiment_epoch(i, v, hz)
+
+    stimulus_end_ix = stim_end_idx
+    response_end_ix = expt_end_ix
 
     post_stim_response_duration = (response_end_ix - stimulus_end_ix) / hz
     completed_expt_epoch = post_stim_response_duration > POSTSTIM_STABILITY_EPOCH
 
     long_response = response_end_ix/hz>LONG_RESPONSE_DURATION
```
njmei: Change to `long_response = (response_end_ix / hz) > LONG_RESPONSE_DURATION` for better readability.

Also, although hz is likely a float, I feel like `from __future__ import division` should be considered since all of this is running on py2.7 right?

ipfx/qc_features.py
```diff
@@ -371,23 +371,33 @@ def current_clamp_sweep_qc_features(sweep_data,sweep_info,ontology):
...
 
     if completed_expt_epoch or long_response:
-        sweep_completion = True
+        return True
     else:
-        sweep_completion = False
+        return False
 
     return sweep_completion
```
njmei: This looks like it never happens so should be removed?

ipfx/epochs.py
```diff
@@ -0,0 +1,132 @@
...
+
+    Returns
+    -------
+    start_index, duration
+    """
+
+    di = np.diff(stim)
+    idxs = np.flatnonzero(di)
+    idxs = idxs[idxs >= idx0]
+
```
nikahk: You should use variables below instead of different return statements.

ipfx/qc_features.py
nikahk: Agree with Nick

tests/test_epochs.py
```diff
@@ -0,0 +1,29 @@
+import ipfx.epochs as ep
+
+
+def test_find_stim_window():
```
nikahk: You should only test one case per test function.

ipfx/epochs.py
```diff
@@ -0,0 +1,132 @@
...
+    elif num_steps <= 0:
+        return 0, 0
+    assert num_steps > 0, "Number of steps should be a positive integer"
+
+    return stim_start-1-num_steps, stim_start-1
+
+
+def get_recording_end_idx(v):
+
+    end_idx = np.nonzero(v)[0][-1]  # last non-zero index along the only dimension=0.
```
njmei: If 'v' has only one dimension then you could just do `end_idx = np.flatnonzero(v)[-1]` and do away with the addition `[0]` indexing...

Same for line 58.

tests/test_epochs.py
njmei: Agree, although you could keep this single function and use [pytest parameterization](https://docs.pytest.org/en/latest/example/parametrize.html) instead...

ipfx/epochs.py
sgratiy: this is a legacy function that I just moved to a different file. I hope to eliminate it altogether soon.

tests/test_epochs.py
sgratiy: I have changed to use parameterization in get_stim_characteristics function. The find_stim_window function may soon go away, so I did not touch the test

sgratiy:
Thanks guys for a careful review. Very helpful.

PR #152, opened 2019-01-18
Fearture/x nwb dat converter one file per group

t-b:


ipfx/bin/run_x_to_nwb_conversion.py
```diff
@@ -7,7 +7,7 @@
 from ipfx.x_to_nwb.DatConverter import DatConverter
 
 
-def convert(inFileOrFolder, overwrite=False, fileType=None, outputMetadata=False, outputFeedbackChannel=False):
+def convert(inFileOrFolder, overwrite=False, fileType=None, outputMetadata=False, outputFeedbackChannel=False, oneFilePerGroup=False):
```
dyf: Can you default this to True?

t-b: @dyf Done.

t-b:
@dyf Done, now also in the main conversion script. I've renamed the flag to multipleGroupsPerFile and that now always defaults to False.

dyf:
Thanks @t-b.  Just to make sure -- the default behavior is that each NWB is all recordings from single cell now, right?

t-b:
@dyf Yes, sorry for the boolean confusion.

PR #153, opened 2019-01-19
Moving epoch indexes from nwb reader into ephys data set

sgratiy:
But also correcting a module import bug.
Using tmpdir_factory to save test nwb files into a temp directory


tests/test_hbg_dataset.py
```diff
@@ -23,19 +20,6 @@ def ontology():
                              [('name', 'extpexpend stimulus'), ('code', 'extpexpend')]
                              ])
 
-
```
njmei: I actually like that these were fixtures. You can use two techniques to make things better:

1. [Fixture finalization and teardown](https://docs.pytest.org/en/latest/fixture.html#fixture-finalization-executing-teardown-code). Basically os.rm() the output_filepath as the finalizer.

2. [Fixture indirect parameterization](https://stackoverflow.com/a/33879151). This way you can pass whatever files as parameters to the fixtures that you'd like.

sgratiy: Thanks for the links, this is good info.
Are you suggesting using fixtures as an alternative to saving nwb into a tmp dir?

sgratiy: Ok. I was not sure myself if I chose a better approach. These two functions do not have return statements thus it seemed odd to me to use fixtures that are being passed as parameters to the test functions.

t-b: How about the following:
- Teach the fixtures to save into a temporary directory
- And let them return the full path to the NWB files

The idea behind the fixtures is that multiple test cases can easily reference them with the smallest amount of code duplication.

tests/conftest.py
```diff
@@ -18,14 +19,22 @@ def load_array_from_zip_file(zip_file, file_name):
...
+
 @pytest.fixture()
 def spike_test_high_init_dvdt():
     return load_array_from_zip_file("data/spike_test_high_init_dvdt.txt.zip", "spike_test_high_init_dvdt.txt")
 
+
+@pytest.fixture(scope="session")
+def tmp_nwb_dir(tmpdir_factory):
+    return tmpdir_factory.mktemp("test_nwb")
+
```
t-b: By using a unique temporary directory for every session we will always redownload the files or?

tests/test_hbg_dataset.py
sgratiy: Yes, this sounds like a good idea. I will change code back to using fixtures, although I am not sure how will it reduce code duplication from the current implementation. Perhaps I am missing something.

t-b: @sgratiy You don't need to duplicate code for the second test case which uses these downloaded files. I think currently we only have one test case for each file.

tests/conftest.py
sgratiy: I believe it does. I am not really happy with this myself. Any suggestions for improvements?

njmei: Maybe this might be helpful? https://github.com/omarkohl/pytest-datafiles

Didn't realize that your test files would be really big.

t-b: How about just downloading the files into the `tests/data` folder?

t-b: And we already do that for the commit files as in `os.path.join(os.path.dirname(__file__), 'data')`.

t-b: @sgratiy Sounds good.

njmei: If you go this route, you should add a check for these files in the relevant tests and have pytest raise an 'expected fail' if they aren't where they should be.

t-b: @sgratiy I've uploaded the files which I added to `tests/data`.

sgratiy: @t-b where have you uploaded them to? I cannot find them on dropbox.

sgratiy: I have changed all nwb tests to use fixtures with parameterization per @njmei  suggestion (thanx). There are no more nwb files in the bin directory. 

I had to do a merge because of some conflicts.

t-b:
@sgratiy @njmei The indirect parameterized fixtures are nice!

PR #154, opened 2019-01-23
X-NWB: Support stimset scale factors for ABF files

t-b:
This also includes the protocol name nicification we discussed earlier ("RAMP_IN1" -> "RAMP").


t-b:
@sgratiy @dyf I've fixed the conflicts, ready for review.

t-b:
@sgratiy Thanks for the hint. I'm usually running `flake8` on the changes, but this did not output errors  related to function naming. Maybe we should set that in the repo global flake8 config file?

sgratiy:
I have not used flake8 yet. Would changes in the flake8 configuration affect me in any way?
By the way, what is the practical benefit of using flake8 for you above PEP8? I use PyCharm which has inspection tools that catch things like function naming conventions.

t-b:
@sgratiy Using flake8 we could do that in a CI job so that every PR is checked. This is fairly common in e.g. https://github.com/NeurodataWithoutBorders/pynwb and has the added benefit that it is editor agnostic (as I'm for example using vim for coding).

PR #155, opened 2019-01-28
Empty stim trace

sgratiy:
The first 4 commits is a refactorying: 
Splitting some modules in order to reduce size and make code more modular.

The last 2 commits are to fix a bug:
For sweeps truncated by the MIES  before stimulus start, the sweep constructor was getting an error. It was happening because sweep creation depended on the proper return from the get_exprimental epoch() in order to align time.  Solution: more aligning the time out of the sweep constructor into a method which is called at a later feature extraction stage.


ipfx/feature_record.py
```diff
@@ -0,0 +1,136 @@
...
+    base = cell_features["long_squares"]["rheobase_sweep"]["spikes"][0]
+    ephys_features["upstroke_downstroke_ratio_long_square"] = nan_get(base, "upstroke_downstroke_ratio")
+    ephys_features["peak_v_long_square"] = nan_get(base, "peak_v")
+    ephys_features["peak_t_long_square"] = nan_get(base, "peak_t")
+    ephys_features["trough_v_long_square"] = nan_get(base, "trough_v")
+    ephys_features["trough_t_long_square"] = nan_get(base, "trough_t")
+    ephys_features["fast_trough_v_long_square"] = nan_get(base, "fast_trough_v")
+    ephys_features["fast_trough_t_long_square"] = nan_get(base, "fast_trough_t")
+    ephys_features["slow_trough_v_long_square"] = nan_get(base, "slow_trough_v")
+    ephys_features["slow_trough_t_long_square"] = nan_get(base, "slow_trough_t")
```
njmei: Seems a bit repetitive maybe instead you could do something similar to:
```
features = ["upstroke_downstroke_ratio", "peak_v", ...]

postfix = '_long_square'
for ft in features:
    new_ft_name = '{}{}'.format(ft, postfix)
    ephys_features[new_ft_name] = nan_get(base, ft)
```

njmei:
Maybe might be worthwhile to write a test or two to make sure no future regressions occur?

ipfx/feature_record.py
```diff
@@ -0,0 +1,136 @@
+import pandas as pd
+import numpy as np
+
+
+def build_cell_feature_record(cell_features, sweep_features):
```
nikahk: This function is doing multiple things. You should split the work.

ipfx/feature_record.py
sgratiy: @njmei Good suggestion. Done.

ipfx/data_set_features.py
```diff
@@ -123,18 +124,19 @@ def extract_sweep_features(data_set, sweep_table):
         sweep_numbers = sorted(sweep_numbers)
         logging.debug("%s:%s" % (stimulus_name, ','.join(map(str, sweep_numbers))))
 
-        sweep_set = data_set.sweep_set(sweep_numbers)
+#        sweep_set = data_set.sweep_set(sweep_numbers)
```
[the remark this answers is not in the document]
sgratiy: @nikahk Done

ipfx/feature_record.py
sgratiy: I refactored a bit to reduce repetition.

sgratiy:
@t-b I agree, I need to add more tests. I will add them on the next PR (i promise :)). I would like to merge this PR asap though to enable processing cells which were previously failing.

ipfx/feature_record.py
sgratiy: @nikahk Thanx for pointing out the "multiple thing" thing. I have refactored to stay close to single responsibility.

PR #156, opened 2019-01-29
Miscellaneous DAT fixes

### LINKED ISSUE #140: ABF/DAT Converter Feedback
(Quoted text is from Jim Berg, rest is my answers)

# General

Tested data: https://www.dropbox.com/sh/o6bkjwnmjtfpodc/AAD3JN1bUfGmV-GnXpMwljhLa?dl=0

> Also - another minor note - I was wondering if it would be possible to generate
> a csv or some other table of all of the metadata values per sweep for a given
> experiment. I don't think we need this for all cells, but if I could have it
> for a few, I could do a sanity check that the values / fields look reasonable.

Yes we could add a NWB to JSON tool which outputs everything except the raw
data. Or we could add the missing metadata to the PDF annotations. Depends on
your use case (i.e. if the output needs to be machine readable or not). If CSV
is really wanted and not only serves as $randomPlaintextFormat we can do that
as well.

-> Postponed

# ABF

> In general, looks good. Roughly, the bottom right plot should look like the
> plot on the top + the bias current. There is an additional offset of ~10 pA -
> I'm not sure where that is coming from, but I'm not so worried about it.

- [x] Scaling factors will be added to metadata JSON file
   - [x] Add support into `mcc_get_settings.py` and document
   - [x] Add support into `ABFConverter.py` for reading the scaling factor and applying it

> For some stimuli, the time base looks slightly off for the top (reconstructed)
> stimulus trace. See Sweep 10000 in H18.29.135.11.41.02.pdf. I'm not sure if
> that is a plotting thing or a data thing.

- [x] Yes that looks odd. I'll investigate. (Note to self: Add source ABF file into NWB and PDF annotation)

> The Triple stim seem to have a more serious error in their reconstructed
> pulses. see sweep 12001 in the above file. It looks like the first part of the
> trace is represented fine, but the actual data traces have much longer rest
> periods at the end.

- [x] Yes, that is a bug.

# DAT

> The only issue that I came up with is the top channel (command) for each trace
> is always off by a factor of 1000.
>
> For H18.28.026.11.09: Sweep 30101, top should be Voltage [V] (bottom is fine)
> Sweep 30401, top should be Current [A] (bottom is fine)
>
> (It happens for all sweeps, I'm just pointing out 2 specific examples).

- [x] Yes that looks odd. The raw NWB data has the same properties. I'll investigate.

> One question I had is whether the bridge balance and bias current settings are
> available for the nwbs that are converted from dat files. They are displayed
> for the pdf from abf files, but not on the one from dat. If they are available,
> it would be great to see them.

- [x] I'll check with Gabor from Tamas Lab to fine tune the DAT file amplifier settings retrieval.

> As far as splitting the dat-nwb files into thier component cells. The one thing
> that looks to be consistent is the first character in the sweep number.
> For H18.28.026.11.09, the first cell is Sweep 10101, second is 20101, third is 30101...
> Somewhat confusingly, for H18.28.028.11.05, first cell is Sweep 1001001, second is 2001001
> My impression is the the second and third group of numbers is some sort of
> organization, like sets and repeats, where in one case they are accounting for
> up to 99 repeats and the next they are accounting for up to 999.
> One way to verify that you are in a new cell is that cells always begin with
> a descr = extpinbath. Relying on this seems error prone though.

You nearly reverse engineered the numbering scheme ;). I'm combining the
Group/Series/Sweep/Trace numbers to get a unique number with fixed length. See
https://github.com/AllenInstitute/ipfx/blob/61143fe04ec487c591a49968d94caa8fac46b473/ipfx/x_to_nwb/utils.py#L68.

So looking at the DAT file in PatchMaster shows that every Group has extpinbath
as first stimulus. So we want to split on a group-by-group basis. Is the group
name (for H18.28.11.05.dat the four groups are named
PGS4_181017_701_A/PGS4_181017_702_A/E-10/E-10 something we should to record in the
NWB or let that be part of the resulting NWB file name?

- [x] Add option to split on group basis, add group name to NWB metadata.

- [x] TB: Remove channel suffix from protocol name or tell them to not add it.

- [x] Clamp mode assignment is wrong, see https://u01-human-brain.slack.com/archives/GC8C4S6SW/p1548366069009800.

t-b:
The clamp mode detection was incorrect and is now fixed.
In addition some TimeSeries properties are now more accurately read out.

Close #140.

### DIFF
--- a/ipfx/x_to_nwb/DatConverter.py
+++ b/ipfx/x_to_nwb/DatConverter.py
@@ -51,7 +51,7 @@ class DatConverter:
             device = self._createDevice()
             nwbFile.add_device(device)
 
-            self.electrodeDict = self._generateElectrodeDict(elem)
+            self.electrodeDict = DatConverter._generateElectrodeDict(elem)
             electrodes = self._createElectrodes(device)
             nwbFile.add_ic_electrode(electrodes)
 
@@ -80,7 +80,8 @@ class DatConverter:
         with Bundle(inFile) as bundle:
             bundle._all_info(root + ".txt")
 
-    def _getClampMode(self, ampState, trace):
+    @staticmethod
+    def _getClampMode(ampState, trace):
         """
         Return the clamp mode of the given amplifier state node.
         """
@@ -123,7 +124,8 @@ class DatConverter:
 
         return counter
 
-    def _generateElectrodeKey(self, trace):
+    @staticmethod
+    def _generateElectrodeKey(trace):
 
         # Using LinkDAChannel and SourceChannel here as these look correct
 
@@ -132,7 +134,8 @@ class DatConverter:
 
         return f"{DAC}_{ADC}"
 
-    def _generateElectrodeDict(self, groups):
+    @staticmethod
+    def _generateElectrodeDict(groups):
         """
         Generate a dictionary of all electrodes in the file.
         Use self._generateElectrodeKey(trace) for generating the key, the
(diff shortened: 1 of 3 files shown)


t-b:
These are the PDF files from the last test, regenerated with this PR:

[H18.28.028.11.05.pdf](https://github.com/AllenInstitute/ipfx/files/2809693/H18.28.028.11.05.pdf)
[H18.28.026.11.09.pdf](https://github.com/AllenInstitute/ipfx/files/2809694/H18.28.026.11.09.pdf)

ipfx/x_to_nwb/DatConverter.py
```diff
@@ -183,6 +182,25 @@ def _convertTimestamp(self, heka_elapsed_seconds):
     def _isValidAmplifierState(ampState):
         return len(ampState.StateVersion) > 0
 
+    @staticmethod
+    def _getAmplifierState(bundle, series, trace_index):
+        """ Return the amplifier state object taking into account different PatchMaster versions """
```
sgratiy: Could you please add a more detailed docstring (describe parameters, returns) to this function and perhaps some other functions to help better understand them.

ipfx/x_to_nwb/DatConverter.py
```diff
@@ -80,32 +80,31 @@ def outputMetadata(inFile):
         with Bundle(inFile) as bundle:
             bundle._all_info(root + ".txt")
 
-    def _getClampMode(self, node):
+    def _getClampMode(self, ampState, trace):
         """
-        Return the clamp mode of the given node.
+        Return the clamp mode of the given amplifier state node.
```
sgratiy: Please add a more detailed docstring

t-b:
@sgratiy @dyf Updated. I've also included a fix for the bridge_balance value in IC due to the answer from heka support.

PR #160, opened 2019-02-05
Redefine the meaning of experimental epoch

sgratiy:
Now experimental epoch is defined as: 
start index: 0.5 second before stimulus start index
end index:  0.5 second after stimulus end index

Thus, experimental epoch is an expected interval of the response.

The new definition simplifies checking for the sweep truncation: 
if the last index of the response is less than the end of the experimental epoch, then experiment is not fully recorded (i.e, truncated).

### DIFF
--- a/ipfx/epochs.py
+++ b/ipfx/epochs.py
@@ -13,18 +13,21 @@ LONG_RESPONSE_DURATION = 5  # this will count long ramps as completed
 
 def get_last_vm_epoch(idx1, hz):
     """
-    Get epoch lasting LAST_STABILITY_EPOCH before the end of recording
+    Get epoch lasting LAST_STABILITY_EPOCH before idx1
 
     Parameters
     ----------
-    idx1
-    hz
+    idx1    : int last index of the epoch
+    hz      : float sampling rate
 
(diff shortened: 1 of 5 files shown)


ipfx/epochs.py
```diff
@@ -97,7 +97,7 @@ def get_stim_epoch_B(i,test_pulse=True):
     return start_idx, end_idx
 
 
-def get_experiment_epoch(i,v,hz):
+def get_experiment_epoch(i,hz):
     """
     Find index range for the experiment epoch.
     The start index of the experiment epoch is defined as stim_start_idx - PRESTIM_DURATION*sampling_rate
```
njmei: It would be nice if many of these function docstrings had descriptions for the parameters. (i.e. like the docstring you have on [line 117](https://github.com/AllenInstitute/ipfx/pull/160/commits/d450a2d8f83929907abb760477811e5f8bbad0e7#diff-70ef283edf9491810282fa9533af83e2R117))

sgratiy: Thanks, I have added docstings to some of the functions that I expect to stay long term.

PR #170, opened 2019-02-14
Enabling selection of epochs on the existing sweep

sgratiy:
Previously, selecting epochs of a sweep required re-loading data from the nwb file.
Now, sweep is created with all data and epochs are detected. We can now request different epochs on the existing sweep without needing to reload data. 
Epochs are now actions of the Sweep class rather than entire data set.
This PR is a bit of WIP...

### DIFF
--- a/ipfx/aibs_data_set.py
+++ b/ipfx/aibs_data_set.py
@@ -8,10 +8,12 @@ import ipfx.nwb_reader as nwb_reader
 
 
 class AibsDataSet(EphysDataSet):
-    def __init__(self, sweep_info=None, nwb_file=None, h5_file=None, ontology=None, api_sweeps=True):
+    def __init__(self, sweep_info=None, nwb_file=None, h5_file=None,
+                 ontology=None, api_sweeps=True, validate_stim=True):
         super(AibsDataSet, self).__init__(ontology)
 
         self.nwb_data = nwb_reader.create_nwb_reader(nwb_file)
+        self.validate_stim = validate_stim
 
         if sweep_info is not None:
             sweep_info = self.modify_api_sweep_info(
@@ -90,7 +92,7 @@ class AibsDataSet(EphysDataSet):
             sweep_record["stimulus_code_ext"] = stim_code_ext
             sweep_record["stimulus_code"] = stim_code
 
-            if self.ontology:
+            if self.validate_stim:
                 sweep_record["stimulus_name"] = self.get_stimulus_name(stim_code)
 
             sweep_props.append(sweep_record)
--- a/ipfx/aibs_data_set.py
+++ b/ipfx/aibs_data_set.py
@@ -91,9 +91,7 @@ class AibsDataSet(EphysDataSet):
 
             sweep_record["stimulus_code_ext"] = stim_code_ext
             sweep_record["stimulus_code"] = stim_code
-
-            if self.validate_stim:
-                sweep_record["stimulus_name"] = self.get_stimulus_name(stim_code)
+            sweep_record["stimulus_name"] = self.get_stimulus_name(stim_code)
 
             sweep_props.append(sweep_record)
 
--- a/ipfx/data_set_features.py
+++ b/ipfx/data_set_features.py
@@ -217,13 +217,6 @@ def extract_cell_features(data_set,
     return cell_features
 
 
-def stim_align_sweep_set(sweep_set):
-
-    for sweep in sweep_set.sweeps:
-        expt_start_idx, _ = ep.get_experiment_epoch(sweep.i,sweep.sampling_rate)
-        sweep.set_time_zero_to_index(expt_start_idx)
-
-
 def select_subthreshold_min_amplitude(stim_amps, decimals=0):
     """Find the min delta between amplitudes of coarse long square sweeps.  Includes failed sweeps.
 
--- a/ipfx/data_set_features.py
+++ b/ipfx/data_set_features.py
@@ -127,8 +127,8 @@ def extract_sweep_features(data_set, sweep_table):
         logging.debug("%s:%s" % (stimulus_name, ','.join(map(str, sweep_numbers))))
 
         sweep_set = data_set.sweep_set(sweep_numbers)
-        sweep_set.select_epoch("response")
-        stim_align_sweep_set(sweep_set)
+        sweep_set.select_epoch("recording")
+        sweep_set.align_to_start_of_epoch("experiment")
 
         dp = detection_parameters(stimulus_name).copy()
         for k in [ "start", "end" ]:
@@ -158,11 +158,10 @@ def extract_cell_features(data_set,
         raise er.FeatureError("No long_square sweeps available for feature extraction")
 
     lsq_sweeps = data_set.sweep_set(long_square_sweep_numbers)
-    lsq_sweeps.select_epoch("response")
-    stim_align_sweep_set(lsq_sweeps)
+    lsq_sweeps.select_epoch("recording")
+    lsq_sweeps.align_to_start_of_epoch("experiment")
 
     lsq_start, lsq_dur, _, _, _ = stf.get_stim_characteristics(lsq_sweeps.sweeps[0].i, lsq_sweeps.sweeps[0].t)
-
 
     logging.info("Analyzing Long Square")
     logging.info("Long square stim start: %f, duration: %f", lsq_start, lsq_dur)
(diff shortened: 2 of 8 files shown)


ipfx/data_set_features.py
```diff
@@ -124,7 +126,9 @@ def extract_sweep_features(data_set, sweep_table):
         sweep_numbers = sorted(sweep_numbers)
         logging.debug("%s:%s" % (stimulus_name, ','.join(map(str, sweep_numbers))))
 
-        sweep_set = data_set.stim_aligned_sweep_set(sweep_numbers)
+        sweep_set = data_set.sweep_set(sweep_numbers)
+        sweep_set.select_epoch("response")
+        stim_align_sweep_set(sweep_set)
```
njmei: It looks like this pattern of:

```
x_set = data_set.sweep_set()
x_set.select_epoch('response')
stim_align_sweep_set(x_set)
```
is repeated for all the changes in this commit. Consider creating a function to avoid repetition.

ipfx/ephys_data_set.py
```diff
@@ -115,20 +114,22 @@ def sweep(self, sweep_number):
...
+        t = np.arange(0, len(sweep_data['stimulus'])) * dt
+        assert len(sweep_data['stimulus']) == len(sweep_data['response'])
 
-        response = sweep_data['response'][sweep_start_ix:sweep_end_ix+1]
-        stimulus = sweep_data['stimulus'][sweep_start_ix:sweep_end_ix+1]
-        t = np.arange(sweep_start_ix, sweep_end_ix + 1) * dt
+        try:
+            epochs = sweep_data['epochs']
+        except KeyError:
+            epochs = {}
```
njmei: I think you should be able to replace 120-123 with just `epochs = sweep_data.get('epochs', dict)`

ipfx/ephys_data_set.py
```diff
@@ -115,20 +114,22 @@ def sweep(self, sweep_number):
         sweep_meta_data = self.get_sweep_meta_data(sweep_number)
         sampling_rate = sweep_data['sampling_rate']
         dt = 1. / sampling_rate
-        sweep_start_ix, sweep_end_ix = ep.get_sweep_epoch(sweep_data['response'])
+        t = np.arange(0, len(sweep_data['stimulus'])) * dt
+        assert len(sweep_data['stimulus']) == len(sweep_data['response'])
```
njmei: Are there cases where `len(sweep_data['stimulus']) != len(sweep_data['response'])`? If so, it should probably be handled? If not, is there any point to the assertion?

ipfx/ephys_data_set.py
sgratiy: good idea

ipfx/ephys_data_set.py
sgratiy: I removed this check altogether for now.

ipfx/sweep.py
```diff
@@ -9,7 +9,12 @@ def __init__(self, t, v, i, clamp_mode, sampling_rate, sweep_number=None, epochs
         self.sampling_rate = sampling_rate
         self.sweep_number = sweep_number
         self.clamp_mode = clamp_mode
-        self.epochs = epochs
+
+        if epochs:
+            self.epochs = epochs
+        else:
+            self.epochs = {}
+
```
njmei: Maybe can be a one-liner: `self.epochs = epochs if epochs else dict`

I don't remember if it should be `dict` or `{}` though...

ipfx/plot_qc_figures.py
```diff
@@ -145,7 +146,7 @@ def plot_single_ap_values(data_set, sweep_numbers, lims_features, sweep_features
             if nspikes:
                 plt.xlim(spikes[0]["threshold_t"] - 0.002, spikes[0]["fast_trough_t"] + 0.01)
         elif type_name == "short_square":
-            plt.xlim(stim_start - 0.002, stim_start + stim_dur + 0.01)
+            plt.xlim(stim_start_shifted - 0.002, stim_start_shifted + stim_dur + 0.01)
```
njmei: What do the values `0.002` and `0.01` (same for the `1e-3` a few lines above) represent? Maybe consider making these `CONSTANT_VARIABLES`?

sgratiy: I agree. Moreover, this entire module needs serious help. I will take care of it as a separate issue in the future.

ipfx/sweep.py
```diff
@@ -2,21 +2,94 @@
...
+        return self._i[start_idx:end_idx+1]
+
+    def select_epoch(self, epoch_name):
+        self.selected_epoch_name = epoch_name
+
+    def set_time_zero_to(self, time_step):
+        dt = 1. / self.sampling_rate
+        self._t = self._t - time_step*dt
+
+    def detect_missing_epochs(self):
```
dyf: rename to `detect_epochs`?

ipfx/sweep.py
```diff
@@ -2,21 +2,94 @@
...
+        """
+        Detect response epoch defined as interval from start to the last non-zero value of the response
+
+        Returns
+        -------
+        start,end: int indices of the epoch
+        """
+        return 0, np.flatnonzero(self.response)[-1]
+
+    def detect_experiment_epoch(self):
```
dyf: are you going to define this?

ipfx/data_set_features.py
```diff
@@ -207,6 +218,13 @@ def extract_cell_features(data_set,
     return cell_features
 
 
+def stim_align_sweep_set(sweep_set):
```
dyf: should this be a method on `SweepSet`?

ipfx/data_set_features.py
```diff
@@ -207,6 +218,13 @@ def extract_cell_features(data_set,
...
+
+    for sweep in sweep_set.sweeps:
+        expt_start_idx, _ = ep.get_experiment_epoch(sweep.i,sweep.sampling_rate)
```
dyf: if you detect the experiment epoch, you could have this function take the epoch name as an argument, as opposed to detecting a specific epoch here.

ipfx/aibs_data_set.py
```diff
@@ -90,7 +92,7 @@ def extract_sweep_meta_data(self):
             sweep_record["stimulus_code_ext"] = stim_code_ext
             sweep_record["stimulus_code"] = stim_code
 
-            if self.ontology:
+            if self.validate_stim:
```
dyf: I think you want to move the validation into `EphysDataSet.get_stimulus_name`.  The way this is, the "name" column will only be populated if you are validating.  

Expected behavior to me:
1) `validate_stim = True`: populate "name" column of stim table but raise an exception if name is not in the ontology
2) `validate_stim = False`: populate "name" column of stim table but raise a warning if name is not in the ontology, leave "name" as None if name is not found.

ipfx/sweep.py
```diff
@@ -2,21 +2,94 @@
...
+
+    def detect_missing_epochs(self):
+        """
+        Detect epochs if they are not provided in the constructor
+
+        """
+
+        epoch_detectors = {
+            "sweep": self.detect_sweep_epoch(),
+            "response": self.detect_response_epoch(),
```
dyf: let's call this "recording"

ipfx/ephys_data_set.py
```diff
@@ -195,6 +195,13 @@ def get_stimulus_name(self, stim_code):
...
-        stim = self.ontology.find_one(stim_code, tag_type="code")
-        return stim.tags(tag_type="name")[0][-1]
-
+        try:
+            stim = self.ontology.find_one(stim_code, tag_type="code")
+            return stim.tags(tag_type="name")[0][-1]
+
+        except KeyError:
+            if self.validate_stim:
+                raise
```
njmei: You should probably add an informative exception message to raise..

ipfx/sweep.py
sgratiy: Good suggestions. I believe, I addressed all of your suggestions. Thanks.

ipfx/ephys_data_set.py
sgratiy: The message is already raised by the calling function in the try block. So I thought it should be enough.

njmei: I think `raise`ing like that breaks the stack trace though?

See: https://stackoverflow.com/questions/18188563/how-to-re-raise-an-exception-in-nested-try-except-blocks

ipfx/data_set_features.py
```diff
@@ -207,6 +217,13 @@ def extract_cell_features(data_set,
     return cell_features
 
 
+def stim_align_sweep_set(sweep_set):
```
dyf: want to get rid of this function now?

PR #176, opened 2019-02-23
Add script for plotting dataset

sgratiy:
The main commit is "add plotting functionality" that enables compactly displaying all data in the dataset.


ipfx/bin/generate_se_input.py
```diff
@@ -14,11 +13,12 @@ def generate_se_input(cell_dir,
...
-        se_input['input_nwb_file'] = lq.get_input_nwb_file(specimen_id)
-        se_input['input_h5_file'] = lq.get_input_h5_file(specimen_id)
+        input_nwb_file = lq.get_input_nwb_file(specimen_id)
+        input_h5_file = lq.get_input_h5_file(specimen_id)
+        if input_h5_file:
+            se_input['input_h5_file'] = input_h5_file
 
-    elif input_nwb_file:
-        se_input['input_nwb_file'] = input_nwb_file
+    se_input['input_nwb_file'] = input_nwb_file
```
t-b: The `elif input_nwb_file:` is still needed as both `specimen_id` and `input_nwb_file` can both be None.

ipfx/bin/plot_ephys_nwb.py
```diff
@@ -0,0 +1,117 @@
...
+
+    height_ratios, width_ratios = axes_ratios(sweep_table)
+
+    fig, ax = plt.subplots(len(height_ratios), 3,
+                           figsize=(15, 7),
+                           gridspec_kw={'height_ratios': height_ratios, 'width_ratios': width_ratios}
+                           )
+
+    i_ax = 0
+    for stimulus_code, sweep_set_table in sweep_table.groupby("stimulus_code"):
```
t-b: Can't you use `enumerate` (https://docs.python.org/2/library/functions.html#enumerate) here?

ipfx/bin/plot_ephys_nwb.py
```diff
@@ -0,0 +1,117 @@
...
+    ax.spines['top'].set_visible(False)
+    ax.spines['right'].set_visible(False)
+    ax.spines['left'].set_visible(False)
+
+
+def get_vertical_offset(data):
+
+    offset = np.max(np.abs(data)) * 1.2
+
+    return offset
```
t-b: Drop the offset variable as it is not required. The function name already tells the reader that this returns an offset.

ipfx/bin/plot_ephys_nwb.py
```diff
@@ -0,0 +1,117 @@
...
+
+def main():
+
+    """
+    # Usage:
+    $ python plot_ephys_nwb_file.py NWB_FILE_NAME
+
+    """
+    if len(sys.argv) == 1:
+        sys.argv.append(DEFAULT_NWB_FILE_NAME)
```
t-b: I'm not a big fan of these kind of default values. Especially as that only works with special setups. E.g. neither on windows nor outside the allen institute LAN.

ipfx/hbg_dataset.py
```diff
@@ -7,10 +7,10 @@
...
 class HBGDataSet(EphysDataSet):
-    def __init__(self, sweep_info=None, nwb_file=None, ontology=None, api_sweeps=True):
+    def __init__(self, sweep_info=None, nwb_file=None, ontology=None, api_sweeps=True, validate_stim=True):
         super(HBGDataSet, self).__init__(ontology)
         self.nwb_data = nwb_reader.create_nwb_reader(nwb_file)
-
+        self.validate_stim = validate_stim
         if sweep_info is not None:
             sweep_info = self.modify_api_sweep_info(
                 sweep_info) if api_sweeps else sweep_info
```
t-b: I'm not understanding the changes here. Currently only `EphysDataSet` uses self.validate_stim but that still does not set it. So if you want to add support for it in the derived classes, you can add support for it in the EphysDataset ctor and hand it down.

In the current  version `validate_stim` is passed into AIBSDataSet and HBGDataSet but not used.

ipfx/bin/plot_ephys_nwb.py
njmei: Agree, definitely use enumerate over manually incrementing `i_ax`.

ipfx/bin/plot_ephys_nwb.py
```diff
@@ -0,0 +1,117 @@
...
+    mng.resize(*mng.window.maxsize())
+    plt.subplots_adjust(left=0.01, right=0.98, bottom=0.02,top=0.92)
+
+
+def axes_ratios(sweep_table):
+
+    height_ratios = []
+    width_ratios = [1,3,3]
+
+    for name, sweep_set_table in sweep_table.groupby("stimulus_code"):
```
njmei: If not using the `name` variable, I would set it to `_`. Or maybe better yet change your `sweep_table.groupby("stimulus_code")` so that it only contains `sweep_set_table`.

ipfx/bin/generate_se_input.py
sgratiy: @t-b I have changed to using mutually exclusive arguments with `argparse`. Now, if `specimen_id` or `input_nwb_file` not provided, then the argparse will catch it before getting to this function.

ipfx/hbg_dataset.py
sgratiy: I can see how it could be confusing. Python is too forgiving of a language to allow this kind of code. I have followed your suggestions and set `validate_stim` in `EphysDataSet` ctor. Thanks.

ipfx/bin/plot_ephys_nwb.py
sgratiy: thanks.

Issue #138, opened 2018-12-17
Feature vector creation module


dyf:
@gouwens has code to organize features and waveforms into vectors for classification.  Integrate that code and build an ArgSchema module for creating those features per-cell.

- [x] resolve merge conflicts
- [x] fix regression tests that broke after merge
- [x] test run feature vector creation
- [x] handle dependency on lims_utils from stash
- [x] remove default paths to Nathan's folders
- [x] add tests of feature vector output
- [ ] add unit tests

PR #184, opened 2019-03-05
Feature vector analysis

gouwens:
Code for feature "vector" extraction to create consistent representations that can be used for dimensionality reduction and classification. Addresses #138. 

Includes the library `feature_vector.py` as well as script to run on many cells at a time (`run_feature_vector_extraction.py`). Also includes an additional runner script (`run_feature_collection.py`) that has a similar interface but produces a CSV of features more akin to what's published at celltypes.brain-map.org.


sgratiy:
@gouwens I am working on merging this. I have noticed that `run_feature_collection.py` is missing somehow from the branch https://github.com/AllenInstitute/ipfx/tree/feature_vector/ipfx/bin although it was committed.

sgratiy:
@gouwens, I am following up on our conversation earlier today about the 'spikes_set' output. 
The cell features already include spikes 
https://github.com/AllenInstitute/ipfx/blob/d4d4b647b138af6e2d7274924f7f12e8bdd85f12/ipfx/stimulus_protocol_analysis.py#L32
and could be accessed as cell_features["long_squares"]["spiking_sweeps"]["spikes"]
Could we use the existing above features instead of introducing features["spikes_set"]
If we can, then we would not need to add/remove "spikes_set".

sgratiy:
Probably I makes sense to only review my commits. More generally, this merge was quite painful and I am looking for feedback on how to make this more straight-forward.

ipfx/stimulus_protocol_analysis.py
```diff
@@ -100,6 +100,9 @@ def analyze(self, sweep_set):
 
     def as_dict(self, features, extra_params=None):
         out = features.copy()
+        out.pop("sweeps")
+        out.pop("spikes_set")
```
njmei: Since it doesn't look like you are doing anything with the popped values from the dictionary, I think it's more explicit if you instead do:
```
del out["sweeps"]
del out["spikes_set"]
```

ipfx/stimulus_protocol_analysis.py
```diff
@@ -210,6 +213,7 @@ def analyze_subthreshold(self, sweep_set):
 
     def as_dict(self, features, extra_params=None):
         out = features.copy()
+        out.pop("spikes_set")
```
njmei: `del out["spikes_set"]`

ipfx/stimulus_protocol_analysis.py
```diff
@@ -288,6 +291,8 @@ def analyze(self, sweep_set):
 
     def as_dict(self, features, extra_params=None):
         out = features.copy()
+        out.pop("sweeps")
+        out.pop("spikes_set")
```
njmei:
```
del out["sweeps"]
del out["spikes_set"]
```

ipfx/bin/lims_queries.py
```diff
@@ -79,13 +88,79 @@ def get_sweep_states(specimen_id):
...
+    SQL = """
+        SELECT sp.id FROM specimens sp
+        JOIN ephys_roi_results err ON sp.ephys_roi_result_id = err.id
+        JOIN projects prj ON prj.id = sp.project_id
+        WHERE prj.code = '%s'
+        """ % project
+
+    if passed_only:
+        SQL += "AND err.workflow_state = 'manual_passed'"
+    print SQL
```
njmei: Remove?

tests/test_run_feature_vector.py
```diff
@@ -0,0 +1,57 @@
...
+from ipfx.bin.run_feature_collection import run_feature_collection
+import pandas as pd
+import pytest
+import ipfx.bin.lims_queries as lq
+
+
+TEST_OUTPUT_DIR = "/allen/aibs/informatics/module_test_data/ipfx/test_feature_vector"
+
+if not lq.able_to_connect_to_lims():
+    pytest.skip("cannot connect to LIMS", allow_module_level=True)
```
njmei: I think this is probably okay. 

But if you think you will be adding more lims required tests in the future, it would probably be more extensible if you marked tests that requires lims with `@pytest.mark.requires_lims` and then in your conftest.py apply the skip to to all tests with the `requires_lims` mark if you can't make a connection to lims.

sgratiy: @njmei Good suggestion. For some silly reasons I will add your suggestions as a part of a separate PR today.

PR #185, opened 2019-03-06
Update qc logic

### LINKED ISSUE #175: When current clamp sweeps are not present, do not exit with error. Instead report a fail tag.
Instead report to lims that sweeps are not available.

sgratiy:
1. When current clamp sweeps are not available, do not raise exception, instead add a fail tag. This would allow qc module to complete, but will not proceed to the next module
2. If blowout sweeps are not available, do not create a fail tag because patchseq cells do not have blowout sweep.

Resolves #175


ipfx/qc_protocol.py
```diff
@@ -39,9 +39,43 @@ def qc_experiment(ontology, cell_features, sweep_features, qc_criteria=None):
...
+
+    Returns
+    -------
+    num_passed_sweeps: int
+    num_sweeps: int
+
+    """
+    num_passed_sweeps = 0
+    for ss in sweep_states:
+        if ss["passed"] is True:
```
njmei: I make this mistake a lot too. You can just do `if ss["passed"]:`

ipfx/qc_feature_extractor.py
```diff
@@ -0,0 +1,280 @@
...
+
+    """
+    if manual_values is None:
+        manual_values = {}
+
+    output_data = {}
+    tag_list = []
+    ontology = data_set.ontology
+    # measure blowout voltage
+    try:
```
njmei: I wonder if it's worthwhile to separate each of these `try` and `except` blocks into their own functions? Your comments seems to suggest that there could be a `measure_blowout_voltage`, `measure_electrode_zero`, `measure_clamp_seal`, `measure_input_and_series_resistance` functions.

ipfx/qc_feature_extractor.py
```diff
@@ -0,0 +1,280 @@
...
+                msg = "Using manual initial access resistance"
+                logging.info(msg)
+                tag_list.append(msg)
+    #
+    output_data['input_resistance_mohm'] = ir
+    output_data["initial_access_resistance_mohm"] = sr
+
+    sr_ratio = None # input access resistance ratio
+    if ir is not None and sr is not None:
+        sr_ratio = sr / ir
```
njmei: Do you need to check that `ir` is not zero? Or does that never happen?

ipfx/qc_feature_extractor.py
```diff
@@ -0,0 +1,280 @@
...
+            if sr is not None:
+                msg = "Using manual initial access resistance"
+                logging.info(msg)
+                tag_list.append(msg)
+    #
+    output_data['input_resistance_mohm'] = ir
+    output_data["initial_access_resistance_mohm"] = sr
+
+    sr_ratio = None # input access resistance ratio
+    if ir is not None and sr is not None:
```
njmei: I think this can just be: `if ir and sr:`

ipfx/sweep_props.py
```diff
@@ -64,3 +64,32 @@ def extract_sweep_features_subset(feature_names, sweep_features):
...
+    -------
+    num_passed_sweeps: int
+        number of sweeps passed QC
+    num_sweeps: int
+        number of sweeps QCed
+
+    """
+    num_passed_sweeps = 0
+    for ss in sweep_states:
+        if ss["passed"] is True:
```
njmei: `if ss.get("passed"):`

ipfx/qc_feature_evaluator.py
```diff
@@ -183,8 +183,7 @@ def qc_current_clamp_sweep(ontology, sweep, qc_criteria=None):
 
 def evaluate_blowout(blowout_mv, blowout_mv_min, blowout_mv_max, fail_tags):
```
njmei: You should probably add a docstring describing this new behavior?

ipfx/qc_feature_evaluator.py
```diff
@@ -50,9 +50,14 @@ def qc_experiment(ontology, cell_features, sweep_features, qc_criteria=None):
         logging.warning(msg)
         cell_state["fail_tags"].append(msg)
 
-    return cell_state, sweep_states
+    if cell_state["fail_tags"]:
```
njmei: `if cell_state.get("fail_tags"):`

ipfx/qc_protocol.py
dyf: If "passed" isn't in the dictionary, schema validation has gone wrong.  This should not be permissive.

ipfx/sweep_props.py
dyf: this should also not be permissive, IMO

ipfx/qc_feature_evaluator.py
dyf: same

ipfx/qc_protocol.py
njmei: Ah, I see!

ipfx/qc_feature_extractor.py
sgratiy: I have refactored this block of code  so that each feature is evaluated by the corresponding function. Thanks.

PR #187, opened 2019-03-13
Properly nandle nwb1 file with no sweeps 

### LINKED ISSUE #180: Index out  of bound in get_sweep_epoch (NSS)

### LINKED ISSUE #182: Sweep naming conventin parsing issue with the nwb1_flavor function

sgratiy:
When nwb1 file has no sweeps (i.e, acquisition and stimulus groups do not have datasets) we were getting an error when determining the flavor of the file. I defaulted the flavor to the "Mies" when sweeps are missing.
Correspondingly properly handle empty sweep table. Previously missing table in some cases would loose the header. Now, even an empty table will always have a header.
Resolve #182

Suppressed warning when real/true sweep number is None. -- too many warnings were making it difficult to examine the pytest output.

Correct the definitions of the sweep and recording epochs for the case when response is an empty array (array of zeros)
Resolve #180 

Correct usage of sweep vs. recording epochs in the modules.


ipfx/nwb_reader.py
```diff
@@ -429,15 +431,32 @@ def get_nwb_version(nwb_file):
...
+                sweep_naming_convention = sweep_names[0].split('_')[0]
 
-            if sweep_naming_convention == "Sweep":
-                return "Pipeline"
-            elif sweep_naming_convention == "data":
+                if sweep_naming_convention == "Sweep":
+                    return "Pipeline"
+                elif sweep_naming_convention == "data":
+                    return "Mies"
+            else:
```
t-b: Could you add a test for the empty nwbv1 file? `test_valid_v1_but_unknown_sweep_naming` can be easily adapted.

ipfx/aibs_data_set.py
```diff
@@ -21,7 +21,14 @@ def __init__(self, sweep_info=None, nwb_file=None, h5_file=None,
...
 
-        self.sweep_table = pd.DataFrame.from_records(sweep_info)
+        self.build_sweep_table(sweep_info)
+
+    def build_sweep_table(self,sweep_meta_data):
+
+        if sweep_meta_data:
+            self.sweep_table = pd.DataFrame.from_records(sweep_meta_data)
+        else:
+            self.sweep_table = pd.DataFrame(columns=self.COLUMN_NAMES)
```
t-b: AibsDataSet.build_sweep_table should be in EphysDataSet as HBGDataSet needs the same fix.

ipfx/ephys_data_set.py
```diff
@@ -41,36 +49,37 @@ def filtered_sweep_table(self,
                              ):
 
         st = self.sweep_table
-
         if current_clamp_only:
-            st = st[st[self.STIMULUS_UNITS].isin(self.current_clamp_units)]
+            mask = st[self.STIMULUS_UNITS].isin(self.current_clamp_units)
+            st = st[mask.astype(bool)]
```
t-b: What issue is the `astype` fixing? Can we get a test for that?

ipfx/epochs.py
```diff
@@ -51,23 +51,39 @@ def get_stability_vm_epoch(stim_start, hz):
...
+    """
+    Detect response epoch defined as interval from start to the last non-zero value of the response
+
+    Returns
+    -------
+    start,end: int indices of the epoch
+    """
+
+    if len(np.flatnonzero(response)) == 0:
+        end_idx = 0
```
t-b: This will return a range with one point where I would expect a range with no points or an exception.

ipfx/nwb_reader.py
```diff
@@ -77,9 +77,11 @@ def read_sweep_from_source(source):
...
                 real_sweep_number = timeseries.attrs["sweep_number"]
 
-        if assumed_sweep_number is not None and assumed_sweep_number != real_sweep_number:
-            warnings.warn("Sweep number mismatch (real: {} vs assumed: {})".format
-                          (real_sweep_number, assumed_sweep_number))
+        if assumed_sweep_number is not None and real_sweep_number is not None:
+
+            if assumed_sweep_number != real_sweep_number:
+                warnings.warn("Sweep number mismatch (real: {} vs assumed: {})".format
+                              (real_sweep_number, assumed_sweep_number))
```
t-b: Change looks good although it takes a second to appreciate it.

ipfx/ephys_data_set.py
sgratiy: the `mask` is a Series of either dtype =bool or object. When mask's dtype = object and the filtered dataframe is empty, the header gets lost. Thus, the easy fix was to force mask's type to `bool`. Mostly likely there is a more robust solution to this problem but I have not found it. I will add a test for it as you suggested.

ipfx/epochs.py
sgratiy: if both start and end indices =0, then the array slice will be empty. Thus, the current implementation will return no points, or am I wrong?

t-b: @sgratiy Okay fine. I just read the function comment and thought it is a one point range.

ipfx/ephys_data_set.py
sgratiy: I added the test `test_empty_table_has_header`
if you we do not use `astype(bool)` when filtering for `stimuli`, then the header is gone.

ipfx/nwb_reader.py
sgratiy: Thanks for a careful review. I hope I addressed all of your points.

ipfx/epochs.py
t-b: I'm still no convinced.

Let's assume response is flat zero.

We venture into the if branch and then return `0, 0`.

One of the callers is `current_clamp_sweep_qc_features`
 https://github.com/AllenInstitute/ipfx/blob/09fcd8e15be3b1fe77231f8f614161323dbf3963/ipfx/qc_feature_extractor.py#L291-L320

which throws away the starting index and only uses the ending index. This ending index is then passed into `get_last_vm_epoch` which returns a tuple where the first index is larger than the second which then results in `measure_vm` return `0, 0`.

If that is how it is supposed to be, well good. At least I would not expect that an empty slice has a property of 0.

PS: You can factor our the `np.flatnonzero(response)` so that it is only called once.

sgratiy: This is a good point and it raises the valid question what to return from the function in case response is flat zero. As it stands now, the `current_clamp_sweep_qc_features` are only called for the "completed" sweeps and flat response will not satisfy it.  I am revamping `current_clamp_sweep_qc_features` now for related reason: when the stimulus is flat zero. So this code should improve.
There are two approaches as I can see: 
1. Check a number of conditions whether the sweeps has all proper recording, stimulus, not truncated early, etc. Fail sweeps  which do not satisfy those conditions and only determine qc features on the passing sweeps. 
2. Go right ahead into determing qc features `current_clamp_sweep_qc_features` , but then use lots of try-except blocks to handle issues with bad data.

Any preference?

t-b: I would prefer 1. This is usually also the way we do things in MIES.

sgratiy: This is my preference as well. I am working on implementing it now. 
Is there something you still need me to update in this PR?

PR #192, opened 2019-03-20
Sweep integrity checks

sgratiy:
The main theme of this PR is to  check integrity of the sweep before computing QC features.
This allows for multiple checks. Currently checking for:
1. whether recording was truncated before completion of experimental epoch
2. whether stimulus is absent
Issues are saved into sweep tags which could be saved into lims.

### DIFF
--- a/ipfx/epochs.py
+++ b/ipfx/epochs.py
@@ -8,10 +8,21 @@ import time_series_utils as tsu
 NOISE_EPOCH = 0.0015
 PRESTIM_STABILITY_EPOCH = 0.5
 POSTSTIM_STABILITY_EPOCH = 0.5
-LONG_RESPONSE_DURATION = 5  # this will count long ramps as completed
 
 
-def get_last_vm_epoch(idx1, hz):
+def get_first_stability_epoch(stim_start_idx, hz):
+
+    num_steps = int(PRESTIM_STABILITY_EPOCH * hz)
+    if num_steps > stim_start_idx-1:
+        num_steps = stim_start_idx-1
+    elif num_steps <= 0:
+        return 0, 0
+    assert num_steps > 0, "Number of steps should be a positive integer"
+
+    return stim_start_idx-1-num_steps, stim_start_idx-1
+
+
+def get_last_stability_epoch(idx1, hz):
     """
     Get epoch lasting LAST_STABILITY_EPOCH before idx1
 
@@ -30,25 +41,14 @@ def get_last_vm_epoch(idx1, hz):
     return idx0, idx1
 
 
-def get_first_vm_noise_epoch(idx, hz):
+def get_first_noise_epoch(idx, hz):
 
     return idx, idx + int(NOISE_EPOCH * hz)
 
 
-def get_last_vm_noise_epoch(idx1, hz):
+def get_last_noise_epoch(idx1, hz):
 
     return idx1-int(NOISE_EPOCH * hz), idx1
-
-
-def get_stability_vm_epoch(stim_start, hz):
-    num_steps = int(PRESTIM_STABILITY_EPOCH * hz)
-    if num_steps > stim_start-1:
-        num_steps = stim_start-1
-    elif num_steps <= 0:
-        return 0, 0
-    assert num_steps > 0, "Number of steps should be a positive integer"
-
-    return stim_start-1-num_steps, stim_start-1
 
 
 def get_recording_epoch(response):
--- a/ipfx/epochs.py
+++ b/ipfx/epochs.py
@@ -166,37 +166,3 @@ def get_experiment_epoch(i,hz):
         return None, None
 
 
-def find_stim_window(stim, idx0=0):
-    """
-    Find the index of the first nonzero positive or negative jump in an array and the number of timesteps until the last such jump.
-
-    Parameters
-    ----------
-    stim: np.ndarray
-        Array to be searched
-
-    idx0: int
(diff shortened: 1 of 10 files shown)


ipfx/qc_features.py
```diff
@@ -33,8 +33,16 @@ def measure_vm(vals):
...
     mean = np.mean(vals)
     rms = np.sqrt(np.mean(np.square(vals-mean)))
-    return mean, rms
-########################################################################
+    return float(mean), float(rms)
+
+
+def measure_vm_delta(mean_start, mean_end):
+
+    if mean_end is not None:
```
njmei: Just wanted to double check, it looks like `mean_end` can either be an `int` or `None`?

sgratiy: It is `float` or `None`

ipfx/epochs.py
```diff
@@ -30,27 +41,16 @@ def get_last_vm_epoch(idx1, hz):
     return idx0, idx1
 
 
-def get_first_vm_noise_epoch(idx, hz):
+def get_first_noise_epoch(idx, hz):
 
     return idx, idx + int(NOISE_EPOCH * hz)
```
njmei: Again another double check `int()` will always truncate towards zero, is that the 'right' behavior?

ipfx/epochs.py
```diff
@@ -86,53 +86,59 @@ def get_sweep_epoch(response):
...
     di_idx = np.flatnonzero(di)  # != 0)
 
     if len(di_idx) == 0:
         return None, None
 
-    if len(di_idx) >= 4:
-        idx = 2  # skip the first up/down assuming that there is a test pulse
+    elif len(di_idx) == 1:
+        raise Exception("Cannot detect and end to the test pulse")
     else:
```
t-b: Drop the else as this allows you to reduce indentation.

ipfx/epochs.py
```diff
@@ -86,53 +86,59 @@ def get_sweep_epoch(response):
...
 
 
-def get_stim_epoch_B(i):
-    """
-    Identify start index, and end index of a general stimulus. Assume there is test pulse.
-    """
+def get_test_epoch(i):
 
     di = np.diff(i)
     di_idx = np.flatnonzero(di)  # != 0)
```
t-b: drop the commented out code fragment as you are touching this function

ipfx/epochs.py
```diff
@@ -86,53 +86,59 @@ def get_sweep_epoch(response):
...
     end_idx = di_idx[-1]
 
     return start_idx, end_idx
 
 
-def get_stim_epoch_B(i):
-    """
-    Identify start index, and end index of a general stimulus. Assume there is test pulse.
-    """
+def get_test_epoch(i):
```
t-b: Does that return the testpulse epoch? Maybe add a comment.

ipfx/qc_features.py
```diff
@@ -3,15 +3,15 @@
...
+    return np.mean(v[idx0:])
 
 
 def measure_electrode_0(curr, hz, t=0.005):
 
     n_time_steps = int(t * hz)
     # electrode 0 is the average current reading with zero voltage input
     # (ie, the equivalent of resting potential in current-clamp mode)
-    return 1e12 * np.mean(curr[0:n_time_steps]*1e-12)
+    return np.mean(curr[0:n_time_steps])
```
t-b: This change and the one in measure_blowout change the exponent from -x to +x. Is that on purpose? If yes can we have a comment for the functions in which units they return their values.

ipfx/epochs.py
```diff
@@ -8,10 +8,21 @@
...
 
-def get_last_vm_epoch(idx1, hz):
+def get_first_stability_epoch(stim_start_idx, hz):
+
+    num_steps = int(PRESTIM_STABILITY_EPOCH * hz)
+    if num_steps > stim_start_idx-1:
+        num_steps = stim_start_idx-1
+    elif num_steps <= 0:
+        return 0, 0
+    assert num_steps > 0, "Number of steps should be a positive integer"
```
t-b: Now I'm all for defensive programming but under what condition should this assertion trigger? The `elif` already catches the `<= 0` ones or?

A function comment would also be nice.

t-b:
- The tests dont' pass here, e.g. ` py.test.exe -v test_epochs.py` fails. Do they pass on your end?

ipfx/qc_features.py
sgratiy: Here the result should remains unchanged. Thus, I removed unnecessary conversion.  Looks like we were converting to pA->A, compute mean and then back again to pA . I am not sure why, perhaps @dyf knows.

ipfx/epochs.py
sgratiy: Not sure what are you asking here...

ipfx/epochs.py
sgratiy: Yes, looks like the assertion would not trigger. I did not write or intend to touch this function. I removed the assert statement, but this function needs to be modified more generally. So I will leave it as a later issue. Thx

ipfx/epochs.py
sgratiy: I added a docstring

ipfx/epochs.py
t-b: @sgratiy `  # != 0)` is commented out code and I think can be removed.

sgratiy: Done

PR #194, opened 2019-03-22
X-NWB: Various minor fixes

t-b:


t-b:
- [ ] Output a better error message for missing protocol files for ABF

ipfx/x_to_nwb/ABFConverter.py
```diff
@@ -172,7 +172,7 @@ def _check(self, abf):
 
                 if np.isnan(abf.sweepC).any():
                     raise ValueError(f"Found at least one 'Not a Number' "
-                                     "entry in stimulus channel {channel} of sweep {sweep}.")
+                                     f"entry in stimulus channel {channel} of sweep {sweep}.")
```
sgratiy: As far as I understand this gets raised when protocol is not found.  I would suggest to make the error message for missing protocol more explicit.

t-b: @sgratiy Could you check the latest version? I added the protocol name and the name of the ABF file.

PR #197, opened 2019-03-26
Enable checking for clamp mode when returning sweep number

### LINKED ISSUE #181: Truncated breakin sweep. Handle exception without failing module.

sgratiy:
Some cells have duplicated test sweeps but marked as VoltageClamp
This feature enable choosing only the desired clamp modes.
Resolves #181

### DIFF
--- a/ipfx/ephys_data_set.py
+++ b/ipfx/ephys_data_set.py
@@ -32,6 +32,9 @@ class EphysDataSet(object):
     SHORT_SQUARE = 'short_square'
     RAMP = 'ramp'
 
+    VOLTAGE_CLAMP = "VoltageClamp"
+    CURRENT_CLAMP = "CurrentClamp"
+
     current_clamp_units = ('Amps', 'pA')
 
     def __init__(self, ontology=None, validate_stim=True):
@@ -46,8 +49,8 @@ class EphysDataSet(object):
         else:
             self.sweep_table = pd.DataFrame(columns=self.COLUMN_NAMES)
 
-    def filtered_sweep_table(self,
-                             current_clamp_only=False,
+    def filtered_sweep_table_ori(self,
+                             clamp_mode=None,
                              passing_only=False,
                              stimuli=None,
                              exclude_search=False,
@@ -56,8 +59,9 @@ class EphysDataSet(object):
                              ):
 
         st = self.sweep_table
-        if current_clamp_only:
-            mask = st[self.STIMULUS_UNITS].isin(self.current_clamp_units)
+
+        if clamp_mode:
+            mask = st[self.CLAMP_MODE] == clamp_mode
             st = st[mask.astype(bool)]
 
         if passing_only:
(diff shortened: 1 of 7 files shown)


tests/test_ephys_data_set.py
```diff
@@ -49,22 +49,22 @@ def test_emtpy_table_has_header():
     assert list(ds.filtered_sweep_table(sweep_number=7)) == list(ds.sweep_table)
 
 
-def test_get_sweep_number_by_stimulus_name_invalid_sweep():
+def test_get_sweep_number_invalid_sweep():
 
     with pytest.raises(IndexError):
         ds = get_dataset()
-        ds.get_sweep_number_by_stimulus_names(['I_DONT_EXIST'])
+        ds.get_sweep_number(['I_DONT_EXIST'])
```
njmei: Don't you need some tests for `get_sweep_number` where `clamp_mode` is not the default value of `None`?

ipfx/ephys_data_set.py
```diff
@@ -81,17 +81,19 @@ def filtered_sweep_table(self,
...
-    def get_sweep_number_by_stimulus_names(self, stimulus_names):
+    def get_sweep_number(self, stimuli, clamp_mode=None):
+
+        sweeps = self.filtered_sweep_table(stimuli=stimuli).sort_values(by=self.SWEEP_NUMBER)
 
-        sweeps = self.filtered_sweep_table(
-            stimuli=stimulus_names).sort_values(by=self.SWEEP_NUMBER)
+        if clamp_mode:
+            mask = sweeps[self.CLAMP_MODE] == clamp_mode
+            sweeps = sweeps[mask]
```
t-b: @sgratiy If you teach `filtered_sweep_table` to allow selection by clamp mode you can pass in both the clamp mode and the simuli and get the result without having to reimplement the logic here.

It would be also nice to check if the given clamp mode exists to detect typos. Or we have, even better, global constants for these strings.

tests/test_ephys_data_set.py
t-b: Yes a test would be good.

ipfx/ephys_data_set.py
sgratiy: This are good suggestions.  Previously, I have stopped short of doing this because I was hesitant of touching `filtered_sweet_table` function used in several places in the code. But, it turned out not difficult in the end. Thnx.

t-b:
@sgratiy Is that ready to look at again? `filtered_sweep_table_ori` vs `filtered_sweep_table` looks like a left-over from development.

sgratiy:
Yes @t-b this is ready to review again. Thanks for pointing out the left over `filtered_sweep_table_ori`. I will take it out tomorrow. In the mean time feel free to review.

t-b:
@sgratiy `filtered_sweep_table_ori` is still in there.

t-b:
@sgratiy I've removed `filtered_sweep_table_ori` in my PR.

Issue #201, opened 2019-04-05
Use real sweep number when getting sweep data


sgratiy:
For nwb_v1 the sweep_numbers were used to be inferred from the acquisition dataset names. Such inferred/assumed sweep number will diverge from the (real) sweep number if sweep is re-acquired.

PR #202, opened 2019-04-05
Deal with re-acquired sweeps

### LINKED ISSUE #200: Handle data with re-acquired sweeps
If sweeps are re-acquired, then use acquisition with the latest starting time

sgratiy:
Resolve #200, #201

### DIFF
--- a/ipfx/aibs_data_set.py
+++ b/ipfx/aibs_data_set.py
@@ -14,85 +14,97 @@ class AibsDataSet(EphysDataSet):
         self.nwb_data = nwb_reader.create_nwb_reader(nwb_file)
 
         if sweep_info is not None:
-            sweep_info = self.modify_api_sweep_info(
-                sweep_info) if api_sweeps else sweep_info
+            sweep_info = self.modify_api_sweep_info(sweep_info) if api_sweeps else sweep_info
         else:
             self.notebook = lab_notebook_reader.create_lab_notebook_reader(nwb_file, h5_file)
             sweep_info = self.extract_sweep_meta_data()
 
         self.build_sweep_table(sweep_info)
 
-
     def extract_sweep_meta_data(self):
         """
 
         Returns
         -------
-        sweep_props: list of dicts
+        sweep_meta_data: list of dicts
             where each dict includes sweep properties
         """
-        sweep_props = []
-        # use same output strategy as h5-nwb converter
-        # pick the sampling rate from the first iclamp sweep
-        # TODO: figure this out for multipatch
+        sweep_meta_data = []
+
         for sweep_name in self.nwb_data.get_sweep_names():
             sweep_record = {}
-            attrs = self.nwb_data.get_sweep_attrs(sweep_name)
-            ancestry = attrs["ancestry"]
-            sweep_record['clamp_mode'] = ancestry[-1]
             sweep_num = self.nwb_data.get_sweep_number(sweep_name)
+
+            sweep_record['starting_time'] = self.nwb_data.get_starting_time(sweep_name)
             sweep_record['sweep_number'] = sweep_num
-            stim_code = self.nwb_data.get_stim_code(sweep_name)
-            if not stim_code:
-                stim_code = self.notebook.get_value("Stim Wave Name", sweep_num, "")
-                logging.debug("Reading stim_code from Labnotebook")
-                if len(stim_code) == 0:
-                    raise Exception(
-                        "Could not read stimulus wave name from lab notebook")
-
-            # stim units are based on timeseries type
-            if "CurrentClamp" in ancestry[-1]:
-                sweep_record['stimulus_units'] = 'pA'
-                sweep_record['clamp_mode'] = 'CurrentClamp'
-            elif "VoltageClamp" in ancestry[-1]:
-                sweep_record['stimulus_units'] = 'mV'
-                sweep_record['clamp_mode'] = 'VoltageClamp'
-            else:
-                # it's probably OK to skip this sweep and put a 'continue'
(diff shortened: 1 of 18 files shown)


sgratiy:
The first 4 commits have already been merged in the previous PR, but they are again here as a part of this PR when merging feature_branch1. This PR is from feature_branch2 made off the feature_branch1. Due to conflicts I had to merge master into feature_branch2 before pushing. Any idea, how I could have avoided getting the first 4 commits as part of this PR?

sgratiy:
Feel free to shred this PR into pieces. This was a hard one. In particular all the merge conflicts result from me not using clamp_mode column, whereas @t-b now assumes that we have this column in the sweep_table. However, during feature_extraction stage, the input json does not provide clamp_mode, thus we either should go without, or extract it again for each sweep. 
To avoid making this PR even more ugly, I have decided to deal with the clamp_mode in a separate issue. #203.

t-b:
@sgratiy Did you start developing this branch and master changed in between? Looks like it. You can avoid having merge commits in the history here if you rebase your branch against master. Something like

```
git checkout master
git pull
git checkout sweep_number
git branch -f bak # copy for going back
git rebase -i master
git push --force
```

should do the trick. This also makes it much easiertor review this PR.

Now regarding the code. I think we first need to think about some general directions.

1.) Do we really need to infer the clamp mode from the simulus units? If you say that the feature extraction stage always defaults to current clamp maybe we can teach `modify_api_sweep_info` to make that decision?

Something like

```
EphysDataSet.CLAMP_MODE: if s.get('clamp_mode') else EphysDataSet.CURRENT_CLAMP,
```

in modify_api_sweep_info. Not that this is really good code. If we want to infer the clamp mode from the stimulus units the easiest solution is just to `git revert` both `af5a1331 (EphysDataSet: Revert introduction of get_clamp_mode, 2019-03-29)` and `c06b7e9a (EphysDataSet: Add clamp mode to the modified sweep info, 2019-03-29)`. I can do that if we agree that this is the best hack.

2.) I would expect that `build_sweep_map` is called implicitly as we currently always want to drop duplicates. Why not just call it in the `NwbReader` constructor?

3.) AibsDataSet/HBGDataSet.extract_sweep_meta_data still work on the sweeps with duplicates. I would expect that all consumers now work on the sweeps without duplicates. The naming is a also a bit dangerous as the functions `get_sweep_names` and `get_sweep_name` have completely different semantics (the latter one never returns duplicates.). How about `get_sweep_names_with_duplicates` for the existing `get_sweep_names` and make a new function `get_sweep_names` which returns the `sweep_name` column from sweep_map_table.

I think deciding 1 is the first thing to do.

sgratiy:
@t-b 
I will try rebasing as you suggested, Thanks. I just have not used interactive rebase yet, and was a bit afraid of dealing with merge conflicts while rebasing individual commits.

sgratiy:
@t-b Thanks for the detailed feedback. It is greatly appreciated
About 1: No we do  not need to infer clamp_mode from the stimulus unit. It is much better if we could grab it from the nwb file. I believe your hack suggestion could work, but I believe I should gets things right with the clamp mode now. Currently, feature extraction only uses CurrentClamp, but it makes me a bit uncomfortable to limit ourselves to the CurrentClamps going forwards.
I spoke with @dyf and he suggested to changes things in a way that would make things more general. This would involve getting clamp_mode if it is not already in the sweep_meta_data. In order to implement this though, I need to modify `extract_sweep_meta_data` to loop over `sweep_number` instead of `sweep_name`. This is because feature_extraction input module has `sweep_number`, but does not have `sweep_name` for identifying sweeps. I can make this changes in this PR while it is fresh in our minds. How does this sound? 

I believe this will take care of your point 3 as well, since we would no longer need `get_sweep_names`.

As for the point 2: I have originally added `add_sweep_map` into the constructor, but then there are some of the nwb_reader tests that are failing because they do no have necessary data to build sweep_map. In order to add `add_sweep_map` into constructor we would need to modify those tests to add necessary data for building sweep_map. Or is there another way?

t-b:
> I spoke with @dyf and he suggested to changes things in a way that would make things more general. This would involve getting clamp_mode if it is not already in the sweep_meta_data. In order to implement this though, I need to modify extract_sweep_meta_data to loop over sweep_number instead of sweep_name. This is because feature_extraction input module has sweep_number, but does not have sweep_name for identifying sweeps. I can make this changes in this PR while it is fresh in our minds. How does this sound?

The change sounds good.

> I believe this will take care of your point 3 as well, since we would no longer need get_sweep_names.

Great.

> As for the point 2: I have originally added add_sweep_map into the constructor, but then there are some of the nwb_reader tests that are failing because they do no have necessary data to build sweep_map. In order to add add_sweep_map into constructor we would need to modify those tests to add necessary data for building sweep_map. Or is there another way?

Which tests exactly?

I think it would be easier for review and also discussing the changes if we split this PR into multiple ones. Maybe one for getting the clamp mode right first?

sgratiy:
>I think it would be easier for review and also discussing the changes if we split this PR into multiple ones. >Maybe one for getting the clamp mode right first?
I can do that, but this means that I would have to abandon this branch for now.

sgratiy:
> Which tests exactly?

test_valid_v1_full_MIES_1
test_valid_v1_full_MIES_2
test_valid_v1_full_MIES_3

t-b:
> By the way, why getting clamp mode right is preferable for you over the the issues with managing the duplicated sweeps?

I just thought from the code in this PR what you need adapted clamp mode support for the sweep table map. If that is not the case you can start with either PR.

PR #204, opened 2019-04-11
Build sweep mapping

### LINKED ISSUE #200: Handle data with re-acquired sweeps

sgratiy:
Resolve #200, #201

### DIFF
--- a/ipfx/nwb_reader.py
+++ b/ipfx/nwb_reader.py
@@ -51,7 +51,6 @@ class NwbReader(object):
         else:
             raise ValueError("Unit {} not recognized from TimeSeries".format(unit))
 
-
     def get_real_sweep_number(self, sweep_name, assumed_sweep_number=None):
         """
         Return the real sweep number for the given sweep_name. Falls back to
@@ -77,18 +76,14 @@ class NwbReader(object):
             elif "sweep_number" in timeseries.attrs:
                 real_sweep_number = timeseries.attrs["sweep_number"]
 
-        if assumed_sweep_number is not None and real_sweep_number is not None:
+            if real_sweep_number is None:
+                warnings.warn("Sweep number not found, returning: None")
 
-            if assumed_sweep_number != real_sweep_number:
-                warnings.warn("Sweep number mismatch (real: {} vs assumed: {})".format
-                              (real_sweep_number, assumed_sweep_number))
-
-        if real_sweep_number is not None:
             return real_sweep_number
-        elif assumed_sweep_number is not None:
-            return assumed_sweep_number
 
         raise ValueError("Could not find a source/sweep_number attribute/dataset.")
+
+
 
     def get_starting_time(self, data_set_name):
         NotImplementedError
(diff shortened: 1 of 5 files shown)


sgratiy:
This is a redo of my previous PR which I am planning to delete.
This PR avoid merge conflict.
I wish I could have added tests for some of the new functions, but this requires either using real nwb files (for which we still have resolved the issue with where to store files) or making nwb test file data on-the-fly.

sgratiy:
I have added a couple of commits related to the proper use of sweep numbers

ipfx/nwb_reader.py
```diff
@@ -90,6 +90,9 @@ def read_sweep_from_source(source):
 
         raise ValueError("Could not find a source/sweep_number attribute/dataset.")
 
+    def get_starting_time(self, data_set_name):
+        NotImplementedError
```
njmei: Should this be `raise NotImplementedError`?

ipfx/nwb_reader.py
```diff
@@ -410,6 +499,14 @@ def get_stim_code(self, sweep_name):
 
         return stim_code
 
+    def get_starting_time(self, data_set_name):
```
njmei: It looks like this function is duplicated twice (once in `class NwbXReader(NwbReader):` and once in `class NwbMiesReader(NwbReader):`). Maybe consider only defining it once in the `NwbReader` class?

t-b: It's not exactly duplicated, in one case it is an array, in the other a plain element. The function `get_scalar_string` (which should then also be renamed as it is not about strings anymore) can be used to make that distinction.

So the NwbReader class could look like:

```
    def get_starting_time(self, data_set_name):
        with h5py.File(self.nwb_file, 'r') as f:
            sweep_ts = f[self.acquisition_path][data_set_name]
            return get_scalar_string(sweep_ts["starting_time"].value)
```

t-b:
@sgratiy 

> I wish I could have added tests for some of the new functions, but this requires either using real nwb files (for which we still have resolved the issue with where to store files) or making nwb test file data on-the-fly.

How about using my FTP for now? Just to keep us going? I'm mailing you the credentials via email.

sgratiy:
@t-b 
We can continue using FTP for now.

t-b:
Overall PR looks really good!

ipfx/nwb_reader.py
sgratiy: @t-b Neat solution. I have adopted it. Thanks for a careful review and for the ftp info.

ipfx/nwb_reader.py
sgratiy: @njmei Good catch, thanks

sgratiy:
@t-b I have added a couple of tests using a nwb with reacquired sweeps and uploded it to the server. I assume this resolves the necessary changes.

PR #205, opened 2019-04-18
Extract clamp mode from nwb file

### LINKED ISSUE #203: Enable getting clamp mode during feature extraction stage
In principle, all those sweeps are CurrentClamp, but perhaps some sanity checking is necessary in order to stay out of trouble. Previously, clamp_mode was inferred from the units. 
Clamp_mode is needed to initialize sweeps. It is also used by filtered_sweep_table, but in principle all the sweeps are CurrentClamp

sgratiy:
At the stage of the feature extraction, the `sweep_record` is missing `clamp_mode` information.
Thus, adding a method for determining the clamp mode. Call it before building a sweep_table.
I had to refactor AibsDataSet and HBGDataSet to factor out individual functions for determining sweep properties.

Resolves #203, #205

### DIFF
--- a/tests/test_aibs_data_set.py
+++ b/tests/test_aibs_data_set.py
@@ -1,0 +1,64 @@
+import pytest
+
+from ipfx.aibs_data_set import AibsDataSet
+from ipfx.aibs_data_set import EphysDataSet
+import ipfx.sweep_props as sp
+from helpers_for_tests import compare_dicts
+
+
+@pytest.mark.parametrize('NWB_file', ['H18.03.315.11.11.01.05.nwb'], indirect=True)
+def test_validate_required_sweep_info(NWB_file):
+
+    sweep_info = [{"sweep_number": 0}]
+    dataset = AibsDataSet(sweep_info, nwb_file=NWB_file, api_sweeps=False)
+
+    assert list(dataset.sweep_table).sort() == dataset.COLUMN_NAMES.sort()
+
+
+def test_modify_sweep_info_keys():
+    d = [{"sweep_number": 123,
+          "stimulus_units": "abcd",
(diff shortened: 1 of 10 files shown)


tests/test_aibs_data_set.py
```diff
@@ -0,0 +1,12 @@
...
+from ipfx.aibs_data_set import AibsDataSet
+
+
+@pytest.mark.parametrize('NWB_file', ['H18.03.315.11.11.01.05.nwb'], indirect=True)
+def test_validate_required_sweep_info(NWB_file):
+
+    sweep_info = [{"sweep_number": 0}]
+    dataset = AibsDataSet(sweep_info, nwb_file=NWB_file, api_sweeps=False)
+
+    print list(dataset.sweep_table) == dataset.FILTER_COLUMN_NAMES
```
njmei: did you mean `assert` instead of `print`?

sgratiy:
@t-b 
>AIBSDataset: get_stimulus_code and get_stimulus_code_ext both use self.notebook but looking at the constructor this is not always available. Do we need to raise an exception if self.notebook is None?

Yeah, this is a problem. I have significantly modified the PR. Now only the clamp_mode is determined inside the EphysDataSet. The other properties are validated by the ArgSchema anyway. 

I have added the tests as you suggested as well.
Thanks for a careful review.

tests/test_aibs_data_set.py
sgratiy: good catch

sgratiy:
@nicain 
I have rewritten that function and dropped validation since it is is already done by ArgSchema. I would need to open a separate issue for doing detailed validation though. Thanks for the tip.

PR #207, opened 2019-04-22
Return a file with the paths of the json files with the extracted features

### LINKED ISSUE #206: Get feature extraction output from specimen_ids
Script requested by a user

sgratiy:
Resolves #206

### DIFF
--- a/ipfx/bin/get_fx_output.py
+++ b/ipfx/bin/get_fx_output.py
@@ -1,0 +1,76 @@
+import allensdk.internal.core.lims_utilities as lu
+import glob
+import os
+import pandas as pd
+import argparse
+
+NO_SPECIMEN = "No_specimen_in_LIMS"
+NO_OUTPUT_FILE = "No_feature_extraction_output"
+
+
+def get_fx_output_json(specimen_id):
+    """
+    Find in LIMS the full path to the json output of the feature extraction module
+    If more than one file exists, then chose the latest version
+
+    Parameters
+    ----------
+    specimen_id
+
+    Returns
+    -------
+    file_name: string
+    """
+
+    query = """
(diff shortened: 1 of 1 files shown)


ipfx/bin/get_fx_output.py
```diff
@@ -0,0 +1,71 @@
...
+    from specimens sp
+    join ephys_roi_results err on err.id = sp.ephys_roi_result_id
+    where sp.id = %d
+    """ % specimen_id
+
+    res = lu.query(query)
+    if res:
+        err_dir = res[0]["storage_directory"]
+
+        file_list = glob.glob(err_dir+'*EPHYS_FEATURE_EXTRACTION_*_output.json')
```
njmei: I would use `os.path.join()` instead of string summation just to be safe.

ipfx/bin/get_fx_output.py
```diff
@@ -0,0 +1,71 @@
...
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--specimen_file', type=str, required=True)
+    parser.add_argument('--specimen_output_json_file', type=str, required=True)
+    args = parser.parse_args()
+
+    return vars(args)
+
+
+def main():
```
njmei: Maybe consider adding this to `setup.py` console scripts? https://python-packaging.readthedocs.io/en/latest/command-line-scripts.html#the-console-scripts-entry-point

sgratiy: @njmei good point. If @abuchin will be using this actively then I will ad it to setup. For now, I will bookmark your link, thanks.

abuchin:
Approved, long time ago...

PR #214, opened 2019-05-16
Enable saving recording date

sgratiy:
Resolve 163

### DIFF
--- a/ipfx/aibs_data_set.py
+++ b/ipfx/aibs_data_set.py
@@ -109,3 +109,6 @@ class AibsDataSet(EphysDataSet):
 
     def get_sweep_data(self, sweep_number):
         return self.nwb_data.get_sweep_data(sweep_number)
+
+    def get_recording_date(self):
+        return self.nwb_data.get_recording_date()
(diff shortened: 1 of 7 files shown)


njmei:
Do you have a test for ` extract_recording_date()`?

sgratiy:
@njmei No, do you think I need it?

njmei:
hmm, it *is* pretty simple so maybe not.

That function does make me wonder though if `"file_create_date"` will always be the name of the field that denotes when a dataset was created? If not, you may have a situation where `extract_recording_date()` returns `None` even if the dataset actually does have a recording date.

sgratiy:
The "file_create_date" is used by NWB 2, but who knows what will happen in the future. Is there a better way to handle this exception?

t-b:
@sgratiy `file_create_date` will stay around for the forseeable future. But it is actually really the file creation/modifcation date. From the naming of the function `extract_recording_date` I'm wondering if you not really want to have the `session_start_time` instead?

As for the code an implementation for `HBGDataset` is missing, it should be same as for `EphysDataSet`.

I'm also wondering if trying to construct a datetime object from the string might be beneficial here.

sgratiy:
@t-b All good points, thanks. I have changed to using `session_start_time` and also constructed datetime_object and only then converted to string. This guarantees that datetime in the json will use a consistent format, unlike in nwb1. Thanks for checking.

PR #235, opened 2019-06-11
Py2to3 conversion

sgratiy:
Converting to Python 3.6
No backward compatibility to python 2.7
From here on will use only Python 3.


ipfx/bin/generate_fx_input.py
```diff
@@ -1,11 +1,12 @@
...
-from run_sweep_extraction import run_sweep_extraction
-from generate_qc_input import generate_qc_input
-from generate_se_input import generate_se_input, parse_args
-from run_qc import run_qc
-import lims_queries as lq
+from .run_sweep_extraction import run_sweep_extraction
+from .generate_qc_input import generate_qc_input
+from .generate_se_input import generate_se_input, parse_args
+from .run_qc import run_qc
+from . import lims_queries as lq
```
njmei: I think relative imports should be avoided if possible.

tests/helpers_for_tests.py
```diff
@@ -21,7 +22,7 @@ def compare_dicts(d_ref, d):
     """
 
     assert sorted(d_ref.keys()) == sorted(d.keys())
-    for k, v in d_ref.items():
+    for k, v in list(d_ref.items()):
```
njmei: pretty sure converting `d_ref.items()` to a `list` is unnecessary?

ipfx/feature_vectors.py
```diff
@@ -306,7 +306,7 @@ def isi_shape(sweep_set, features, n_points=100, min_spike=5):
         end_index = tsu.find_time_index(isi_sweep.t, isi_sweep.t[fast_trough_index] + 0.1)
         isi_raw = isi_sweep.v[fast_trough_index:end_index] - threshold_v
 
-        width = len(isi_raw) / n_points
+        width = len(isi_raw) // n_points
```
njmei: Just as a double check, this *should* be truncating division?

ipfx/nwb_reader.py
```diff
@@ -229,8 +230,8 @@ def get_pipeline_version(self):
                     # keys are even numbered, corresponding values are in
                     #   odd indices
                     for i in range(len(info)):
-                        if info[i] == 'version':
-                            version = info[i+1]
+                        if info[i].decode('utf-8') == 'version':
```
njmei: You should probably use that `to_str()` function here?

ipfx/feature_vectors.py
sgratiy: Yes, this should be an integer division, because the result should be an integer.

PR #240, opened 2019-06-13
New depolarizing subthreshold feature

gouwens:
I've added analysis of subthreshold depolarizing steps to the feature vector code. In doing so, I also adjusted the logic of the other subthreshold (hyperpolarizing) analyses.

There is also some clean-up of QC/error messages, and I'm using the "cached" version of filtered dvdt in more places in spike detection to improve performance (since doing the filtering before dvdt calculation is the most time-intensive part of the code).

Note that one of the reference test files (`fv_subthresh_norm_T301.npy`) needs to be updated for the associated test to pass since it's picking a different sweep to analyze now. We could also add a new npy file for the depolarizing subthreshold feature vector (e.g. `fv_subthresh_depol_norm_T301.npy`).


ipfx/feature_vectors.py
```diff
@@ -97,21 +97,53 @@ def extract_feature_vectors(data_set,
...
+        lsq_sub_sweeps, lsq_sub_features,
+        lsq_sub_start, lsq_sub_end,
+        lsq_supra_spx,
+        target_sampling_rate=target_sampling_rate,
+        ap_window_length=ap_window_length
+    )
     return all_features
 
 
+def feature_vectors_multipatch(lsq_supra_sweeps, lsq_supra_features,
```
sgratiy: This function has a very complex input and does not seem to represent a reusable action.  Writing tests for this function with so many arguments is difficult especially since it accepts a spike extractor object. Thus, I would suggest to eliminate `feature_vectors_multipatch` and place the code inside `extract_multipatch_feature_vector` (the same is for `extract_feature_vector`). Later on I would like us to refactor extract_feature_vector/extract_multipatch_feature_vector to make them more maintainable.

ipfx/bin/run_feature_vector_extraction.py
```diff
@@ -19,20 +19,67 @@ class CollectFeatureVectorParameters(ags.ArgSchema):
...
+    passed_sql = """
+    select swp.sweep_number from ephys_sweeps swp
+    where swp.specimen_id = %s
+    and swp.sweep_number = any(%s)
+    and swp.workflow_state like '%%passed'
+    """
 
+    iclamp_st = data_set.filtered_sweep_table(clamp_mode=data_set.CURRENT_CLAMP, stimuli=stimuli_names)
 
-def data_for_specimen_id(specimen_id, passed_only, ap_window_length=0.005):
```
sgratiy: I wish we did not need to query LIMS here, since this prevents using this functionality by external users. We should be able to modify the QC module to allow selecting the desired sweeps according to the `sweep_qc_option`. We should strive to limit usage of `specimen_id` to only generate top level input (i.e., nwb_file, h5_file), otherwise external users would not be able to use it.

ipfx/feature_vectors.py
```diff
@@ -153,7 +153,7 @@ def feature_vectors(lsq_sweeps, ssq_sweeps, ramp_sweeps,
     """Feature vectors from stimulus set features"""
 
     result = {}
-    result["step_subthresh"] = step_subthreshold(lsq_sweeps, lsq_features, lsq_start, lsq_end)
+    result["step_subthresh"] = step_subthreshold(lsq_sweeps, lsq_features, lsq_start, lsq_end, amp_tolerance=5)
```
sgratiy: I have made a PR with some test examples for `isi_shape`, `step_subthreshold` and `subthreshold_depol_norm`. But I cannot decide if these are good or good enough. Could you please review that PR and modify tests as needed. Also, we need tests for the remaining features.

ipfx/bin/run_feature_vector_extraction.py
```diff
@@ -88,22 +114,30 @@ def data_for_specimen_id(specimen_id, sweep_qc_option, ap_window_length=0.005):
...
+    if (nwb_path is None) or (len(nwb_path) == 0): # could not find an NWB file
         logging.debug("No NWB file for {:d}".format(specimen_id))
         return {"error": {"type": "no_nwb", "details": ""}}
 
     # Check if NWB has lab notebook information, or if additional hdf5 file is needed
     h5_path = None
-    with h5py.File(nwb_path, "r") as h5:
-        if "general/labnotebook" not in h5:
-            h5_path = lq.get_igorh5_path_from_lims(roi_id)
+    try:
```
sgratiy: Getting `h5_path` is a good candidate for a separate function.

sgratiy:
Looking in more detail I realized that to make feature vector functionality maintainable will require some refactoring of the code. Generally, making functions more modular (i.e, where each function does just one thing) and cover them with tests would greatly help.

ipfx/feature_vectors.py
```diff
@@ -97,21 +97,53 @@ def extract_feature_vectors(data_set,
     return all_features
 
 
-def extract_multipatch_feature_vectors(lsq_sweeps, lsq_start, lsq_end,
+def extract_multipatch_feature_vectors(lsq_supra_sweeps, lsq_supra_start, lsq_supra_end,
```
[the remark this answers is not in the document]
gouwens: @sgratiy I’m happy to add tests. I’m not sure if unit tests are the most helpful here, so it would be more of evaluating with test data and comparing to example output. I’m not sure how you have that infrastructure set up. For example, it doesn’t seem like the files for comparison are in the repository?

ipfx/feature_vectors.py
gouwens: I’m not quite sure what you mean about it not being a reusable action - it’d be run for hundreds or thousands of cells, so that’s a repeated set of steps that seems appropriate to encapsulate? Why is a long list of arguments inherently an issue?

But I’m open to refactoring it. If anything, maybe it’s more appropriate to move most of the logic of the `extract_feature_vector()` and `feature_vectors()` functions (and multipatch equivalents) to the runner scripts and just leave the individual feature vector calculation functions in `feature_vectors.py`.

ipfx/bin/run_feature_vector_extraction.py
gouwens: I don’t think I necessarily agree with the broader comment - it’s a runner script, not a library. So some level of specificity for the interface is necessary here.

More specifically, there are a host of issues with the QC status of sweeps in the NWB files that are actually being used, so it’s just not possible to rely on that directly at this time.

I could take the runner scripts out of the library entirely, but it seems like there is some use of having them as examples to learn from.

ipfx/feature_vectors.py
gouwens: I’ll work on the tests (again, I need more information about how to implement them here).

ipfx/bin/run_feature_vector_extraction.py
gouwens: I’m not sure it makes sense for the `h5_path` part to be split from the NWB file loading preceding it, but that whole thing does probably make sense as a separate function.

gouwens:
@sgratiyI have refactored pretty extensively along the lines you suggested. I have moved the functions with complex inputs from `feature_vectors.py` into the runner script (since those are pretty specific to a particular analysis). I refactored the actual feature vector calculation functions into simpler units and created tests for all the major functions there.

I have rewritten the runner script `run_feature_vector_extraction.py` so that there's an option to use NWB files downloaded via Allen SDK, rather than needing LIMS. I also added the option to output everything to one HDF5 file, which is nice and consolidated.

Most of my new tests use synthetic data, so they don't rely on external files. However, one of the tests that you created (for `step_subthreshold`) doesn't pass because the reference file has the wrong data (I think it has the output from `subthresh_depol_norm`), so that needs to be replaced. And the reference files for the test for `run_feature_vector_extraction.py` will need to be updated for the new outputs.

I've also added missing documentation to several functions. I think overall the code is now better organized and should be more maintainable.

One note - I also refactored the "multipatch" version of the runner script, but that relies on a synphys-specific library that doesn't seem to be working with Python 3. It's only an example, so I'm not too worried about it at this moment. It can be updated when that library is updated.

ipfx/feature_vectors.py
sgratiy: We already largely discussed this, and you latest refactoring is looks really good. I just thought I would still response to your questions here for clarity.  
By not being reusable I meant that the function is called in only one place in the code. But, yes you are correct it is certainly reusable because it is being used to process many cells in a loop.
Generally, long list of arguments  makes it harder to understand, maintain and test functions. But, I am not suggesting that you should necessarily strive to reduce the number of arguments in the functions. Trying to make functions doing just "one thing" should naturally reduce number of arguments.

sgratiy:
The latest refactors looks much much better. You split functions and added many tests -- this is great, thanks.
Re: failing tests. I have double checked that the tests for the latest of my commit that you merged into your branch did pass all the tests. Thus, it looks like the tests failed as a result of refactoring. For example, `test_run_feature_vector` fails because of change in the arguments of the `run_feature_vector` function. I can fix those if you like.

gouwens:
@sgratiy I've added tests for the `identify` functions and ones with synthetic data for `isi_shape`. I also updated the argument list for `run_feature_vector_extraction`, but it fails since the files it wants to compare against aren't there yet.

gouwens:
@sgratiy I pulled in your PR for the test fixes and also pulled in the new ontology functionality and updated the related scripts. I'm seeing tests passing on my end (at least the ones relating to feature_vector - I can't run them all because of a deprecation error that I just created an issue for).

PR #242, opened 2019-06-14
Ignore Test pulse when measuring resistance

sgratiy:
Resolves 229

### DIFF
--- a/ipfx/qc_features.py
+++ b/ipfx/qc_features.py
@@ -61,11 +61,7 @@ def get_r_from_stable_pulse_response(v, i, t):
     ir: float input resistance
     """
 
-
-    dv = np.diff(v)
-    up_idx = np.flatnonzero(dv > 0)
-    down_idx = np.flatnonzero(dv < 0)
-    assert len(up_idx) == len(down_idx), "Truncated square pulse"
+    up_idx, down_idx = get_square_pulse_idx(v)
     dt = t[1] - t[0]
     one_ms = int(0.001 / dt)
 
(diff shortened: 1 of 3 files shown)


ipfx/qc_features.py
```diff
@@ -117,3 +111,31 @@ def get_r_from_peak_pulse_response(v, i, t):
...
+    """
+    dv = np.diff(v)
+
+    up_idx = np.flatnonzero(dv > 0)[1:] # skip the very first pulse (test pulse)
+    down_idx = np.flatnonzero(dv < 0)[1:]
+
+    assert len(up_idx) == len(down_idx), "Truncated square pulse"
+
+    for up_ix, down_ix in zip(up_idx, down_idx):
+        assert up_ix < down_ix, "Negative pulse"
```
njmei: Would it be useful to know if there are multiple instances of negative pulses? Or do you only care about whether the pulse trace has any negative pulses?

ipfx/qc_features.py
```diff
@@ -117,3 +111,31 @@ def get_r_from_peak_pulse_response(v, i, t):
...
+    -------
+    up_idx, down_idx: list, list
+        up, down indices
+    """
+    dv = np.diff(v)
+
+    up_idx = np.flatnonzero(dv > 0)[1:] # skip the very first pulse (test pulse)
+    down_idx = np.flatnonzero(dv < 0)[1:]
+
+    assert len(up_idx) == len(down_idx), "Truncated square pulse"
```
njmei: I wonder if it might be better to have [custom exceptions](https://docs.python.org/3/tutorial/errors.html#user-defined-exceptions) because in your tests, you expect an `AssertionError` but the source could come from either a `Truncated square pulse` or a `Negative pulse`.

ipfx/qc_features.py
sgratiy: I only care about the latter. If there is a negative pulse then there is a problem.

ipfx/qc_features.py
sgratiy: What would be the benefit of using custom exception here? Yes, the sources of error are different, but I do not need to handle them differently. If an error is raised then I will see the cause via the assertion statement.

njmei: It matters for your tests, because you are doing `with pytest.raises(AssertionError):` if somewhere down the line, some regression is introduced that causes one assertion to fail but others to be fine, you could accidentally 'pass' a test that should fail.

sgratiy: @njmei I matched the string representation of the exception. I think this will be good enough for such a test. Thanks for the suggestion.

PR #243, opened 2019-06-17
Fail sweeps with no test pulse

sgratiy:
Resolve 230


ipfx/epochs.py
```diff
@@ -132,11 +133,16 @@ def get_test_epoch(i):
     start_idx,end_idx: int tuple
         start,end indices of the epoch
     """
```
njmei: Should probably update docstrings?

PR #250, opened 2019-07-02
Automate ontology

sgratiy:
Resolve 166

### DIFF
--- a/ipfx/bin/run_feature_extraction.py
+++ b/ipfx/bin/run_feature_extraction.py
@@ -47,11 +47,9 @@ def run_feature_extraction(input_nwb_file,
     if len(sweep_info) == 0:
         raise er.FeatureError("There are no QC-passed sweeps available to analyze")
 
-    if stimulus_ontology_file:
-        mso.make_stimulus_ontology_from_lims(stimulus_ontology_file)
-    else:
+    if not stimulus_ontology_file:
         stimulus_ontology_file = StimulusOntology.DEFAULT_STIMULUS_ONTOLOGY_FILE
-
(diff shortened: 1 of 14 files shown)


ipfx/stimulus.py
```diff
@@ -27,6 +26,8 @@ def has_tag(self, tag, tag_type=None):
 
 class StimulusOntology(object):
 
+    DEFAULT_STIMULUS_ONTOLOGY_FILE = os.path.join(os.path.dirname(__file__), 'defaults/stimulus_ontology.json')
```
njmei: I wonder if it might make sense to have a separate config file for these kinds of variables?

ipfx/bin/run_feature_extraction.py
```diff
@@ -47,7 +47,12 @@ def run_feature_extraction(input_nwb_file,
     if len(sweep_info) == 0:
         raise er.FeatureError("There are no QC-passed sweeps available to analyze")
 
-    ont = StimulusOntology(ju.read(stimulus_ontology_file)) if stimulus_ontology_file else StimulusOntology()
+    if stimulus_ontology_file:
+        mso.make_stimulus_ontology_from_lims(stimulus_ontology_file)
+    else:
+        stimulus_ontology_file = StimulusOntology.DEFAULT_STIMULUS_ONTOLOGY_FILE
+
+    ont = StimulusOntology(ju.read(stimulus_ontology_file))
```
njmei: It looks like this block of code lines 50-55 are repeated quite a few times in other modules (`run_qc.py`, `run_sweep_extraction.py`) I would try centralizing this into a function unless you think it might change differently for those other modules.

PR #258, opened 2019-07-17
fix wrong name for certain noise stimuli

gouwens:
Fixes noise stimuli wrongly marked as short squares


gouwens:
I guess if they are gone that still addresses the issue of them being incorrectly included in short square analyses. But it does seem like it would make sense to still keep track of them (even just to avoid generating warnings when loading those NWBs).

bergjim:
I just added them back. Sorry, I thought they were duplicates!

PR #262, opened 2019-07-26
Get data from nwb in base SI units

sgratiy:
Both data and unit names must be in base SI units. 
For example, we should have data in Volts (not mV)

### DIFF
--- a/ipfx/ephys_data_set.py
+++ b/ipfx/ephys_data_set.py
@@ -140,6 +140,7 @@ class EphysDataSet(object):
 
         epochs = sweep_data.get('epochs')
         clamp_mode = sweep_record['clamp_mode']
+
         if clamp_mode == "VoltageClamp":
             v = sweep_data['stimulus']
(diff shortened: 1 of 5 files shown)


gouwens:
Why is this a desired change?

ipfx/ephys_data_set.py
```diff
@@ -149,6 +150,9 @@ def sweep(self, sweep_number):
         else:
             raise Exception("Unable to determine clamp mode for sweep " + sweep_number)
 
+        v *= 1.0e3    # convert units V->mV
```
njmei: If you are getting `sweep_data` from the nwb, it would be a good idea to check that the units of the data are actually in `V` and `A`. At least based on the description of that units field, it could potentially be in `mv` or `pA` already, especially if the data comes from a different source?

sgratiy: Good point, I have added a validation function to check that both stimulus and response use SI units.

PR #275, opened 2019-08-19
Nwb2 regression test

sgratiy:
pynwb started using object_id for objects that changes every time file gets created even from the same data. Thus, using h5diff returns a difference when performing regression testing.
Per suggestion from pynwb developer, I added an option to monkey patched object_id property to produce a fixed value for all object ids. This will be used for regression testing.
By default this option will be disabled and allow generating genuine object_ids when converting files.

### DIFF
--- a/ipfx/bin/run_nwb1_to_nwb2_conversion.py
+++ b/ipfx/bin/run_nwb1_to_nwb2_conversion.py
@@ -4,6 +4,7 @@ import os
 import argschema as args
 from ipfx.x_to_nwb.NWBConverter import NWBConverter
 
+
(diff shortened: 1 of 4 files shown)


njmei:
Hmm... rather than directly modifying the `NWBConverter` class I think it might be a better to create a `TestingNWBConverter` class which inherits from `NWBConverter` and have the patching of the hdmf `Container.object_id` happen before a super() call.

sgratiy:
@njmei I agree, making a separate test class is a better approach. I have update the code accordingly. thanks for the suggestion.

PR #277, opened 2019-08-21
Roundtrip tests for nwb2 files

sgratiy:
Basic round trip testing for writing and then reading time series: acquisition, stimulus


sgratiy:
@njmei Thanks, the build is passing now.

Issue #164, opened 2019-02-05
Pipeline to output NWB2 file


sgratiy:
Scope out required work.
Make epics and issues as appropriate.

~~Request new .dat and .abf files from collaborators for conversion and validation.~~ see #285 

Update Ephys Feature Extraction module to write the output file as NWB2
Write nwb2 only when input is nwb2.
Use pynwb_validate for validation

- [x] read-and-append interface to ipfx nwb2 files. This should implement set_spike_times
- [x] A function which constructs either the above class or allensdk.core.nwb_data_set.NwbDataSet from a path. Logic for inferring nwb version is here.
 - [x] roundtrip test for set_spike_times
- [x] accessor methods for reading spikes (needed for roundtrip tests)

wbwakeman:
http://confluence.corp.alleninstitute.org/display/~sergeyg/NWB2+processing+by+the+IVSCC+ephys+pipeline

NileGraddis:
adding myself per discussion in sprint planning

NileGraddis:
@sgratiy could you tell me if:
1. this is correct
2. completing the work laid out here is sufficient to close this issue.
3. the design seems sensible

### goals

The ipfx pipeline should, if run with NWB2 input data, output an NWB2 files containing extracted features.

### background
The feature extraction stage (fx) runs at the end of the ipfx pipeline. It takes as input an NWB file and, after running successfully, generates a copy of the NWB file that contains detected spike times. The code that does this is here:
https://github.com/AllenInstitute/ipfx/blob/664f9f3578fe9a34f27c80b742b2af91459521c7/ipfx/bin/run_feature_extraction.py#L19

Although fx can read NWB 1 or 2 files using
https://github.com/AllenInstitute/ipfx/blob/664f9f3578fe9a34f27c80b742b2af91459521c7/ipfx/data_set_utils.py#L6
`embed_spike_times` can only append to NWB 1 files.

Tests for NWB2 writing live here:
https://github.com/AllenInstitute/ipfx/blob/664f9f3578fe9a34f27c80b742b2af91459521c7/tests/test_nwb1_nwb2_conversion.py

There are the following existing NWB interfaces:
- [NwbReader](https://github.com/AllenInstitute/ipfx/blob/664f9f3578fe9a34f27c80b742b2af91459521c7/ipfx/nwb_reader.py). NWB1 & NWB2. RO.
[...]

PR #288, opened 2019-10-15
Embedding spikes into nwb file

sgratiy:
Solves #164

Need a feedback for best way of implementing tests: I did both ways using classes and functions
I am getting some warnings when adding data to nwb2 which still needs to be resolved.

### DIFF
--- a/ipfx/bin/run_feature_extraction.py
+++ b/ipfx/bin/run_feature_extraction.py
@@ -8,30 +8,38 @@ from ipfx.stimulus import StimulusOntology
 from ipfx._schemas import FeatureExtractionParameters
 from ipfx.data_set_utils import create_data_set
 import ipfx.sweep_props as sp
-import ipfx.bin.make_stimulus_ontology as mso
 import allensdk.core.json_utilities as ju
-from allensdk.core.nwb_data_set import NwbDataSet
+import ipfx.nwb_append as appender
 
 import ipfx.plot_qc_figures as plotqc
 import ipfx.logging_utils as lu
 
 
-def embed_spike_times(input_nwb_file, output_nwb_file, sweep_features):
-    # embed spike times in NWB file
-    tmp_nwb_file = output_nwb_file + ".tmp"
+def embed_spike_times(input_nwb_file, output_nwb_file, sweep_spikes):
 
+    tmp_nwb_file = output_nwb_file + ".tmp"
     shutil.copy(input_nwb_file, tmp_nwb_file)
-    for sweep_num in sweep_features:
-        spikes = sweep_features[sweep_num]['spikes']
-        spike_times = [ s['threshold_t'] for s in spikes ]
-        NwbDataSet(tmp_nwb_file).set_spike_times(sweep_num, spike_times)
+
+    nwb_data = appender.create_nwb_appender(tmp_nwb_file)
+    nwb_data.add_spike_times(sweep_spikes)
 
     try:
         shutil.move(tmp_nwb_file, output_nwb_file)
     except OSError as e:
         logging.error("Problem renaming file: %s -> %s" % (tmp_nwb_file, output_nwb_file))
         raise e
-    logging.debug("Embedded spike times into output.nwb file")
+    logging.debug(f"Embedded spike times into the {output_nwb_file}")
+
+def collect_spike_times(sweep_features):
+
+    spikes = {}
+
+    for sweep_num in sweep_features:
+        spike_features = sweep_features[sweep_num]['spikes']
+        spike_times = [ s['threshold_t'] for s in spike_features ]
+        spikes[sweep_num] = spike_times
+
+    return spikes
 
 
 def run_feature_extraction(input_nwb_file,
@@ -77,7 +85,8 @@ def run_feature_extraction(input_nwb_file,
         feature_data = {'cell_state': cell_state}
 
     if not cell_state["failed_fx"]:
-        embed_spike_times(input_nwb_file, output_nwb_file, sweep_features)
+        sweep_spike_times = collect_spike_times(sweep_features)
+        embed_spike_times(input_nwb_file, output_nwb_file, sweep_spike_times)
 
         if qc_fig_dir is None:
             logging.info("qc_fig_dir is not provided, will not save figures")
--- a/ipfx/nwb_append.py
+++ b/ipfx/nwb_append.py
@@ -1,0 +1,84 @@
+import os
+import pynwb
+from allensdk.core.nwb_data_set import NwbDataSet
+import ipfx.nwb_reader as nwb_reader
+from pynwb import TimeSeries
+from pynwb import ProcessingModule
+
+
+class NwbAppender(object):
+
+    def __init__(self, nwb_file_name):
+        if os.path.isfile(nwb_file_name):
+            self.nwb_file_name = nwb_file_name
+        else:
+            raise FileNotFoundError(f"Cannot locate {nwb_file_names}")
+
+    def add_spike_times(self, sweep_num, spike_times):
+        raise NotImplementedError
+
+    def set_spike_time(self,sweep_spike_times):
+        raise NotImplementedError
+
+
+class Nwb1Appender(NwbAppender):
+
+    def __init__(self, nwb_file_name):
+        NwbAppender.__init__(self, nwb_file_name)
+        self.nwbfile = NwbDataSet(self.nwb_file_name)
+
+    def add_spike_times(self, sweep_spike_times):
+
+        for sweep_num, spike_times in sweep_spike_times.items():
+            self.nwbfile.set_spike_times(sweep_num, spike_times)
+
+
+class Nwb2Appender(NwbAppender):
+
+    def __init__(self, nwb_file_name):
+        NwbAppender.__init__(self, nwb_file_name)
+
+        io = pynwb.NWBHDF5IO(self.nwb_file_name, 'a')
+        self.nwbfile = io.read()
+        io.close()
+
+    def add_spike_times(self, sweep_spike_times):
+
+        spike_module = ProcessingModule(name='spikes',
+                                        description='detected spikes')
+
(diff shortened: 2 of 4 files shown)


ipfx/bin/run_feature_extraction.py
```diff
@@ -8,15 +8,15 @@
...
 import allensdk.core.json_utilities as ju
 from allensdk.core.nwb_data_set import NwbDataSet
+import ipfx.nwb_writer as writer
 
 import ipfx.plot_qc_figures as plotqc
 import ipfx.logging_utils as lu
 
 
-def embed_spike_times(input_nwb_file, output_nwb_file, sweep_features):
+def embed_spike_times_old(input_nwb_file, output_nwb_file, sweep_features):
```
njmei: If this is going to be `old` maybe might be good idea to add a deprecation warning? I'm guessing there's a reason it can't just be removed?

sgratiy: good catch, I just need to remove the `old` one
Could you please comment on test class implementation?

tests/test_write_nwb.py
```diff
@@ -0,0 +1,140 @@
...
+    make_skeleton_nwb1_file(input_nwb_file_name)
+    embed_spike_times(input_nwb_file_name, output_nwb_file_name, sweep_spike_times)
+
+    nwb_data = nwb_reader.create_nwb_reader(output_nwb_file_name)
+
+    for sweep_num, spike_times in sweep_spike_times.items():
+        assert np.allclose(nwb_data.get_spike_times(sweep_num), spike_times)
+
+
+class ForTestEmbedSpikeTimesNwb(object):
```
njmei: Maybe name this `TestEmbedSpikeTimesNwbBase` instead?

ipfx/nwb_writer.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def __init__(self, nwb_file_name):
+        NwbWriter.__init__(self, nwb_file_name)
+        self.nwbfile = NwbDataSet(self.nwb_file_name)
+
+    def add_spike_times(self, sweep_num, spike_times):
+        self.nwbfile.set_spike_times(sweep_num, spike_times)
+
+    def save(self):
+        pass
```
njmei: Does nwb1writer not need to save?

ipfx/bin/run_feature_extraction.py
```diff
@@ -34,6 +34,37 @@ def embed_spike_times(input_nwb_file, output_nwb_file, sweep_features):
...
+    shutil.copy(input_nwb_file, tmp_nwb_file)
+
+    nwb_data = writer.create_nwb_writer(tmp_nwb_file)
+    for sweep_num,spike_times in sweep_spikes.items():
+        nwb_data.add_spike_times(sweep_num, spike_times)
+
+    nwb_data.save()
+
+    try:
+        shutil.move(tmp_nwb_file, output_nwb_file)
```
njmei: Why is there a need to rename the nwbfile? Wouldn't it be better to specify a output savepath to the `NWBWriter.save()` method?

tests/test_write_nwb.py
```diff
@@ -0,0 +1,140 @@
...
+from ipfx.bin.run_nwb1_to_nwb2_conversion import make_nwb2_file_name
+from hdmf import Container
+import numpy as np
+import ipfx.nwb_reader as nwb_reader
+
+from ipfx.bin.run_feature_extraction import embed_spike_times
+
+
+
+
```
kschelonka: I think the functional implementation is clearer.
I would recommend making use of pytest fixtures and parametrization. Since `test_embed_spike_times_into_nwb2` and `test_embed_spike_times_into_nwb1` are exactly the same outside of arguments and filenames, you can just have one function with the inputs defined using the `pytest.mark.parametrize`  decorator.

tests/test_write_nwb.py
```diff
@@ -0,0 +1,140 @@
...
+    io = pynwb.NWBHDF5IO(nwb2_file_name, 'w')
+    io.write(nwbfile)
+    io.close()
+
+
+
+def test_embed_spike_times_into_nwb2(tmpdir_factory):
+
+    sweep_spike_times = {
```
kschelonka: I would just have this at the module level since it's reused in both tests without changes

ipfx/nwb_writer.py
sgratiy: for nwb1, file operation is handled by the [NwbDataSet](https://github.com/AllenInstitute/AllenSDK/blob/a06728262f82bc92e0865311b7b82891edc11a06/allensdk/core/nwb_data_set.py#L258) class that uses context manager and closes file automatically on exit from context.

tests/test_write_nwb.py
sgratiy: I agree, the function implementation looks clearer. I am just not sure how to implement  this test with a single function since nwb1 and nwb2 files are created by `make_skeleton_nwb1_file` and `make_skeleton_nwb2_file`, respectively. Can I pass different fixtures to the same test function?

tests/test_write_nwb.py
sgratiy: If I do that, then pytest tries to run test function from this class as well, but I only need tests to be run on child classes.

ipfx/bin/run_feature_extraction.py
sgratiy: My understanding is that decision to use an intermediate tmp file was made because nwb1 had issues with corrupting files. Thus, the input nwb file is copied to tmp, spikes were added and if all went well, the output nwb is created. If we bypass the tmp file, then output nwb would be created which may not have spikes, or if it gets corrupted then it would also mess up an output nwb from the previous pipeline run.

ipfx/nwb_writer.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+class Nwb2Writer(NwbWriter):
+
+    def __init__(self, nwb_file_name):
+        NwbWriter.__init__(self, nwb_file_name)
+
+        io = pynwb.NWBHDF5IO(self.nwb_file_name, 'a')
+        self.nwbfile = io.read()
```
NileGraddis: I think this will fail if the nwbfile does not already exist. OK for your embed_spike_times use case, but a bit odd for a class called `Nwb2Writer`

ipfx/nwb_writer.py
```diff
@@ -0,0 +1,71 @@
...
+        io.close()
+
+        self.nwbfile.add_unit_column('sweep_number', description="sweep number")
```
NileGraddis: what happens if there is already a sweep_number column?

ipfx/nwb_writer.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def add_spike_times(self, sweep_num, spike_times):
+        self.nwbfile.add_unit(spike_times=spike_times, sweep_number=sweep_num)
```
NileGraddis: the units table [has a bunch of fields](https://github.com/NeurodataWithoutBorders/pynwb/blob/f62f52e300a098b161ccca7c1c5fd598feef28e5/src/pynwb/misc.py#L160). Are you not using those?

Not saying this is a bad decision, mind.

tests/test_write_nwb.py
NileGraddis: You could write something like:
```
@pytest.mark.parametrize("make_skeleton_file", [make_skeleton_nwb1_file, make_skeleton_nwb2_file])
def test_embed_spike_times(make_skeleton_file):
    ...
    make_skeleton_file(input_nwb_file_name)
    ...
```

Also, the `make_skeleton_nwb\d_file` functions are not actually pytest fixtures.

ipfx/nwb_writer.py
sgratiy: I assume pynwb follows h5py's modes with "a": Read/write if exists, create otherwise (default)

ipfx/nwb_writer.py
sgratiy: Yes, I am not using those.

ipfx/nwb_writer.py
sgratiy: It will error out -- good point. Also this brings me to needing to check if Units table already exists and remove it. Do you know what is the way to delete a container?

ipfx/nwb_writer.py
NileGraddis: I think the `io.read` call will error out if the file does not already exist.

ipfx/nwb_writer.py
NileGraddis: why do you need to replace the table of units? Surely you just want to overwrite (or error out) one line?

tests/test_write_nwb.py
sgratiy: done

ipfx/nwb_writer.py
sgratiy: I changed implementation to using processing module and timeseries. Units table turned out to be a poor fit because it assumes a all spikes for unit saved in a single table row. In my case I need to save row for each sweep for the same unit. I also tried using DynamicTable, but had no luck with saving spike times when sweeps had different number of spikes. @NileGraddis using timeseries was your original suggestion. :) Hope you still like it.

NileGraddis: I have no problem with this implementation (since it seems like there is no official way to resolve this issue).

NileGraddis: This seems to obviate the original comment, but still ... what happens if there are already some spike times in this file?

ipfx/nwb_writer.py
NileGraddis: Is this ok to be append-only?

sgratiy: You are right again, `io.read` errors out if file does not exist.
I turned the class into `NwbAppender` and raise `FineNotFoundError` exception if file does not exist.

ipfx/nwb_writer.py
sgratiy: If time series for the sweep already exists, it will error out.

PR #292, opened 2019-10-17
Chirp feature code and off-pipeline analysis tools

tmchartrand:
This is a bit of an update to Nathan's chirp code:

- actual chirp features (peak frequency, etc.) in addition to Nathan's feature vectors
- transforming single chirp sweeps not just averaged sweeps
- an initial foray into providing some tools for running feature analysis in an off-pipeline context. duplicates some functionality in run_feature_collection, run_feature_vector_extraction, etc but in a more modular form that can be used across different use cases.

@gouwens , in this version I've left your code functionally unchanged, but one point that should be resolved is how to deal with sweeps of different duration (different stimulus codes). Right now you're windowing and averaging before transform, but I'd suggest the appropriate approach is either to restrict to only the final chirp sweep type (other two seem to have been used in early cells to explore parameters), or to average FVs after transform. I could easily link into my code to do this if you want.

### DIFF
--- a/ipfx/bin/run_chirp_feature_collection.py
+++ b/ipfx/bin/run_chirp_feature_collection.py
@@ -1,0 +1,12 @@
+from ipfx.chirp import extract_chirp_features
+import ipfx.offpipeline_utils as op
+import argschema as ags
+
+module = ags.ArgSchemaParser(schema_type=op.FeatureCollectionParameters)
+# TODO: maybe argschema can take care of this logic with a custom field?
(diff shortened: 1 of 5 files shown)


tmchartrand:
@sgratiy  I'm going to close this for now after talking with Nathan. He will integrate some of my updates to the core functionality, then I will reopen an update rebased on that. In the meantime, this code may still be useful as an example of adding in new features if you're considering any broader updates in that direction.

PR #298, opened 2019-10-25
297/circle ci

### LINKED ISSUE #297: Run tests for ipfx in cloud-based CI tool
In order to better support external users of this repository, it makes sense to set up an automated build system that is not limited to internal Allen Institute systems only.

Investigate and implement one of the start tools for this. See AllenSDK tests for example.
~200 tests; about 90% cloud-compatible.
Non-integration tests take ~10 minutes to run

sgratiy:
Resolves #297

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,69 @@
+# Python CircleCI 2.0 configuration file
+#
+# Check https://circleci.com/docs/2.0/language-python/ for more details
+#
+version: 2
+jobs:
+  build:
+    parallelism: 1
+    environment:
+      TEST_INTERNAL: false
+
+    docker:
+      # specify the version you desire here
+      # use `-browsers` prefix for selenium tests, e.g. `3.6.1-browsers`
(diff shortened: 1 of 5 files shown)


sgratiy:
@kschelonka I was not able to get codecov to work. Would be nice to have, but it is not explicitly called for this issue.

.circleci/config.yml
```diff
@@ -0,0 +1,74 @@
...
+
+      - run:
+          name: install test dependencies
+          command: |
+            . venv/bin/activate
+            pip install -r test_requirements.txt
+            pip install codecov
+
+      # run tests!
+      # this example uses Django's built-in test-runner
```
kschelonka: I'd remove this comment since you're using pytest to reduce confusion

kschelonka:
Once the comment from the circleci example is removed I approve

.circleci/config.yml
sgratiy: good catch, I have amended the commits.

PR #301, opened 2019-11-01
274/sweep trailing zeros

sgratiy:
Resolve 274

### DIFF
--- a/ipfx/ephys_data_set.py
+++ b/ipfx/ephys_data_set.py
@@ -2,9 +2,7 @@ import warnings
 import logging
 import pandas as pd
 import numpy as np
-from ipfx.stimulus import StimulusOntology
 from ipfx.sweep import Sweep,SweepSet
-
 
 class EphysDataSet(object):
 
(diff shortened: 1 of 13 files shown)


ipfx/ephys_data_set.py
```diff
@@ -197,17 +195,41 @@ def get_recording_date(self):
...
             'stimulus': np.ndarray,
             'response': np.ndarray,
             'stimulus_unit': string,
             'sampling_rate': float
         }
-
         """
-        raise NotImplementedError
+
+        sweep_data = self.nwb_data.get_sweep_data(sweep_number)
```
njmei: If you are relying on the `self.nwb_data` attribute then you should create something like to indicate that subclasses should specify `nwb_data`:
```
@property
def nwb_data(self):
    raise NotImplementedError
```

sgratiy: Thanks @njmei. I have implemented the property as you suggested

njmei:
LGTM!

PR #320, opened 2020-02-01
Add support for mies nw bv2 files

t-b:
I've not raised the pynwb/hdmf versions in the requirements.txt, as I think this needs to be coordinated on a broader scale.

This is also the reason for the CI failure.


sgratiy:
@t-b after pointing your pr to a new base there still one failing test
test_valid_v2_full_MIES[valid_v2_MIES-NWBv2-0655f7e4.nwb] - tests.test_nwb_reader
It cannot create IntracellularElectrode object
Could you please take a look at circleci build for details?

t-b:
@sgratiy I'm not sure that just using a different base is enough. I've rebased and now the new test is passind.

sgratiy:
@t-b aha, thanks for resolving this. You probably saw that there was another test failing
test_valid_v2_full_ABF[2018_03_20_0005.nwb] - tests.test_nwb_reader
somehow it is getting units as 'amperes'. 
Not sure where this is coming from since previously this test was passing.

t-b:
@sgratiy https://github.com/AllenInstitute/ipfx/pull/333#issuecomment-586034629

sgratiy:
got it. Would the solution be t include "amperes" as one of the valid SI units?

t-b:
> got it. Would the solution be t include "amperes" as one of the valid SI units?

Yes, I'd try that.

t-b:
Closing in favour of #333.

PR #327, opened 2020-02-12
Abf nwb

### LINKED ISSUE #319: x_to_nwb.ABFConverter failed for clamp mode 2
`ABFConverter(src_path, dest_path, outputFeedbackChannel=True)`

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-3-73dd97f2f1b8> in <module>
      7     dest_path = '/'.join(src_path.split('/')[:6] + ['nwb', dest_fname])
      8 
----> 9     ABFConverter(src_path, dest_path, outputFeedbackChannel=True)
     10 
     11 

~/dev/ipfx/ipfx/x_to_nwb/ABFConverter.py in __init__(self, inFileOrFolder, outFile, outputFeedbackChannel, compression)
     79         nwbFile.add_ic_electrode(electrodes)
     80 
---> 81         for i in self._createStimulusSeries(electrodes):
     82             nwbFile.add_stimulus(i)
     83 

~/dev/ipfx/ipfx/x_to_nwb/ABFConverter.py in _createStimulusSeries(self, electrodes)
    391                                              sort_keys=True, indent=4)
    392 
--> 393                     seriesClass = getStimulusSeriesClass(self._getClampMode(abf, channel))
    394 
    395                     stimulus = seriesClass(name=name,

~/dev/ipfx/ipfx/x_to_nwb/conversion_utils.py in getStimulusSeriesClass(clampMode)
     54         return CurrentClampStimulusSeries
     55     else:
---> 56         raise ValueError(f"Unsupported clamp mode {clampMode}.")
     57 
     58 

ValueError: Unsupported clamp mode 2.
```

Looking at 
https://github.com/AllenInstitute/ipfx/blob/050ed97c9590996e099c9e4f7609b62bd7847e8a/ipfx/x_to_nwb/conversion_utils.py#L22-L23

it looks like only modes 0 and 1 are supported, corresponding to voltage and current clamps. What is mode 2? Can we add support for it?

I can see about sharing the data if that is necessary. It's not mine so I will need to get permission first.

fyi @satra, @t-b

bendichter:
Add support for `IZeroClampSeries` (I=0) for Multiclamp 700B amplifiers (mode 2)

fix #319

### DIFF
--- a/ipfx/x_to_nwb/ABFConverter.py
+++ b/ipfx/x_to_nwb/ABFConverter.py
@@ -18,7 +18,7 @@ from pynwb.device import Device
 from pynwb import NWBHDF5IO, NWBFile
 from pynwb.icephys import IntracellularElectrode
 
-from ipfx.x_to_nwb.conversion_utils import PLACEHOLDER, V_CLAMP_MODE, I_CLAMP_MODE, \
+from ipfx.x_to_nwb.conversion_utils import PLACEHOLDER, V_CLAMP_MODE, I_CLAMP_MODE, I0_CLAMP_MODE, \
      parseUnit, getStimulusSeriesClass, getAcquiredSeriesClass, createSeriesName, convertDataset, \
      getPackageInfo, createCycleID
 
@@ -362,7 +362,6 @@ class ABFConverter:
         counter = 0
 
         for file_index, abf in enumerate(self.abfs):
-
             stimulus_description = ABFConverter._getProtocolName(abf.protocol)
             scale_factor = self._getScaleFactor(abf, stimulus_description)
 
@@ -519,6 +518,10 @@ class ABFConverter:
                 d["bias_current"] = np.nan
                 d["bridge_balance"] = np.nan
                 d["capacitance_compensation"] = np.nan
+            elif clampMode == I0_CLAMP_MODE:
+                d["bias_current"] = np.nan
+                d["bridge_balance"] = np.nan
+                d["capacitance_compensation"] = np.nan
             else:
                 warnings.warn("Unsupported clamp mode {clampMode}")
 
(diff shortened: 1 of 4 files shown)


bendichter:
@t-b Can you take a look at this please? The tests appears to be failing for an unrelated issue

bendichter:
@luiztauffer An `IZeroClampSeries` does not require a `CurrentClampSeries`. I see you have written one with all zeros. It's not a big deal, especially since with compression this dataset gets reduced to almost nothing, I think it would be better to remove it, as this pattern of all zeros is implied by the `IZeroClampSeries`.

bendichter:
Also, all of the new data files from the Kriegstein dataset that are zero clamp only have a single sweep. Is that right? It could be right, I just wanted to make sure we were not accidentally omitting other sweeps on these sessions.

luiztauffer:
> @luiztauffer An `IZeroClampSeries` does not require a `CurrentClampSeries`. I see you have written one with all zeros. It's not a big deal, especially since with compression this dataset gets reduced to almost nothing, I think it would be better to remove it, as this pattern of all zeros is implied by the `IZeroClampSeries`.

This is fixed now

luiztauffer:
@bendichter 
1. Metafile is added as optional argument for `run_x_to_nwb_conversion.convert()`  function and `ABFConverter` class.
2. NWBFile is created like before, but now it overwrites the default information with the user provided metadata.
3. Subject can be created if passed by metadata. However, the enforced version of pynwb restricts the use of some fields (`date_of_birth`) because of old bugs

ipfx/x_to_nwb/ABFConverter.py
```diff
@@ -314,35 +313,27 @@ def getFileComments(abfs):
...
+        if 'lab_meta_data' in metadata:
             from ndx_labmetadata_abf import LabMetaData_ext
-
-            lab_metadata = LabMetaData_ext(
-                name=meta_lab_meta_data['name'],
-                cell_id=meta_lab_meta_data['cell_id'],
-                tissue_sample_id=meta_lab_meta_data['tissue_sample_id'],
-            )
-            nwbfile.add_lab_meta_data(lab_metadata)
+            nwbfile.add_lab_meta_data(LabMetaData_ext(**metadata['lab_meta_data']))
```
luiztauffer: 'lab_meta_data' is a key inside NWBFile: 
metadata['NWBFile']['lab_meta_data']

That's why, before constructing NWBFile, we should use:
```pyhton
meta_ini_user = copy.deepcopy(metadata['NWBFile'])
meta_lab_meta_data = meta_ini_user.pop('lab_meta_data', None)
```

t-b:
@bendichter Thanks for the PR. This PR seems to do different things at once. Can you make a separate one for the addition of IZ support? This should be easy to add.

bendichter:
Will do.

Issue #321, opened 2020-02-05
Resolve ephys pipeline output differences between nwb1 and nwb2


sgratiy:
When running pipeline for the same experiment saved in nwb1 vs nwb2 there are some differenced in the metadata produced by the pipeline:
e.g.: nwb2 vs nwb1
"stimulus_code_ext": "X6SP_Rheo_DA_0" vs stimulus_code_ext": "X6SP_Rheo[9]
"bridge_balance_mohm": [ 12.942612648010254 ] vs "bridge_balance_mohm": 12.942613,
same for leak_pa
stimulus_scale_factor value is different between the two:

- [x] use `get_scalar_value` function when reading data from nwb2
- [x] read stimulus_scale_factor from the labnotebook for the MIES nwb2
- [x] resolve value difference in stimulus_scale factor
- [x] Review/revise and possibly redesign the nwb reading classes. Related to the issue #286
- [x] Add test to compare pipeline outputs for nwb2 and nwb1

PR #334, opened 2020-02-14
321/process nwb2

sgratiy:
Address #321


wbwakeman:
If you are interested, I have hundreds of MIES generated NWB2 files at /allen/aibs/technology/waynew/ephys/output/Z/ephys/20200207_nwb2/

ipfx/hbg_dataset.py
```diff
@@ -93,3 +86,15 @@ def get_stimulus_code_ext(self, sweep_num):
 
     def get_recording_date(self):
         return self.nwb_data.get_recording_date()
+
+def get_finite_or_none(d, key):
+
+    try:
+        value = d[key]
+    except KeyError:
```
MatthewAitken: `d.get(key)` is imo a much more succinct and explicit way to do this- get() returns none by default, or you can specify a default value, e.g. `d.get(key, None)`. It is very slightly slower though, so if this is done a whole lot maybe just leave it.

ipfx/nwb_reader.py
```diff
@@ -111,17 +113,17 @@ def get_unit_name(stim_attrs):
     def get_long_unit_name(unit):
         if not unit:
             return "Unknown"
-        if unit.startswith('A'):
+        elif unit.startswith('A') or unit == 'amps':
```
MatthewAitken: is this as specific of a check we need? `startswith` seems possibly ambigious...

sgratiy: I specified unit names explicitly as you suggested. thanks.

PR #337, opened 2020-02-29
tests/test_nwb1_nwb2_conversion.py: Upgrade to newer pynwb versions

t-b:
Pynwb versions since 6b2a628f (Hardcode units of PatchClampSeries
subclasses (#1036), 2019-08-05) use fixed units for the patch clamp
series derived classes.

After changing that the tests pass again.


sgratiy:
@t-b The nwb1_to_nwb2 conversion is not being used. Instead we are using pxp to nwb2 conversion. Thus, it makes sense to remove the nwb1 to nwb2 coversion and corresponding test and close this PR.

sgratiy:
Close per #410

PR #339, opened 2020-03-03
316/nwb2 publish

NileGraddis:
Adds an attach_metadata executable, which can stage metadata in a collection of sinks, then dispatch to external targets. Currently supports out-of-place NWB appending, and a best-guess DANDI sink.

The actual metadata fields are a best guess, since their spec is WIP, but this framework should be sufficiently versatile to accommodate new fields.


ipfx/attach_metadata/__main__.py
```diff
@@ -0,0 +1,146 @@
...
+    output = {}
+    output.update({"inputs": parser.args})
+    output.update(
+        run_attach_metadata(
+            inputs_record["nwb2_sinks"] + inputs_record["dandi_sinks"],
+            inputs_record["metadata"]
+        )
+    )
+
+    print(output)
```
MatthewAitken: do you want to leave this print here?

NileGraddis: in fact, I do not

PR #344, opened 2020-03-05
ABFConverter: Feature/make json settings file truly optional

### LINKED ISSUE #341: Extend ABFConverter.py to not require mcc settings file
The existing ABFConverter script within x_to_nwb requires providing a settings.json file that reflects multiclamp commander amplifier settings. 

As this settings file is typically not available for historical datasets, it would be helpful to modify the ABF conversion script to make this providing of the settings file optional and to set those fields to NA (or some other equivalent).

This issue is related to #261

### LINKED ISSUE #342: Extend ABFConverter to accept user-defined channel names
The current ABFConverter script uses a hard-coded list of expected channel names: 
https://github.com/AllenInstitute/ipfx/blob/74904808979dab54125df23de0d0e27aceba9207/ipfx/x_to_nwb/ABFConverter.py#L31

We need to extend this to meet two additional use cases:

- Occasionally, user defined channel names are used. 
- Occasionally, in some two-electrode set ups, one electrode is not used in the experiment but this channel is not explicitly turned off. We need a way to tell the converter about this so that the data from this channel is not added to the resulting NWB file.

t-b:
@youngseok-seo Please test.

The JSON settings file was never really required, but not passing it in resulted in a bunch of warnings.
You can now pass `no-searchSettingsFile` to avoid the warnings. In addition I've also removed a few other warnings and upgraded to pynwb 1.3.0.

@sgratiy We should discuss if and how we merge PRs from external users.

Close #341.
Close #342.

### DIFF
--- a/ipfx/bin/run_x_to_nwb_conversion.py
+++ b/ipfx/bin/run_x_to_nwb_conversion.py
@@ -91,7 +91,7 @@ def main():
                                help="Helper for debugging which outputs HTML/TXT files with the metadata contents of the files.")
     common_group.add_argument("--log", type=str, help="Log level for debugging, defaults to the root logger's value.")
     common_group.add_argument("filesOrFolders", nargs="+",
-                               help="List of ABF files/folders to convert.")
+                               help="List of files/folders to convert.")
 
     abf_group.add_argument("--protocolDir", type=str,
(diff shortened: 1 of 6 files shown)


t-b:
Closing. This is superseeded by https://github.com/byte-physics/x-to-nwb.

PR #346, opened 2020-03-06
add support for zero current clamp

bendichter:


bendichter:
@t-b here's a PR for just IZ support

t-b:
@sergey What do you think?

bendichter:
@t-b can we move forward with this? did you @ the right name?

PR #348, opened 2020-03-06
286/mies hbg refactor

MatthewAitken:


ipfx/dataset/mies_nwb_data.py
```diff
@@ -0,0 +1,51 @@
...
+
+class MIESNWBData(EphysNWBData):
+    """
+    Provides an Ephys Data Interface to a MIES generated NWB file
+
+    """
+
+    def __init__(self,
+                 nwb_file: str,
+                 lab_notebook_reader: LabNotebookReader,
```
sgratiy: To me it makes more sense to construct labnotebook reader inside the constructor rather than ingesting it. The notebook is created from the same nwb_file as the MIESNWBData object. This prevents someone from injecting a wrong notebook from a different nwb file in the object. To me it seems that the notebook and MIESNWBData should live and die together.

ipfx/dataset/hbg_nwb_data.py
```diff
@@ -0,0 +1,45 @@
...
+
+class HBGNWBData(EphysNWBData):
+    """
+    Provides an Ephys Data Interface to an HBG generated NWB file
+
+    """
+
+    def __init__(self,
+                 nwb_file: str,
+                 ontology: StimulusOntology,
```
sgratiy: I would rather pass the str stimulus_ontology file name not the StimulusOntology object. This look cleaner to me since this parallels the other argument (nwb_file). What are we gaining by passing the ontology object?

ipfx/dataset/ephys_nwb_data.py
```diff
@@ -0,0 +1,388 @@
...
+        return None
+
+    return value
+
+
+class EphysNWBData(EphysDataInterface):
+    """
+    Abstract base class for implementing an EphysDataInterface with an NWB file
+
+    Provides common NWB2 reading and writing functionality
```
sgratiy: A couple of days a ago @NileGraddis and I were thinking that it makes sense to limit the scope of the EphysDataInterface to just reading capabilities to enable feature extraction. Thus, for a user who only cares about extracting features, it would only be necessary to implement the reading interface. In turn, Nile has implemented writing capabilities by directly accessing the pynwb.io object.

PR #350, opened 2020-03-09
286/onto labnb

gnayuy:
refactored stimulus and labnotebook and added unit tests


tests/test_labnotebook.py
```diff
@@ -0,0 +1,40 @@
...
+from builtins import range
+import os
+import pytest
+import h5py
+import numpy as np
+from ipfx.dataset.labnotebook import LabNotebookReaderIgorNwb
+
+TEST_DATA_PATH = os.path.join(os.path.dirname(__file__), 'data')
+
+@pytest.mark.parametrize('NWB_file', ['H18.03.315.11.11.01.05.nwb'], indirect=True)
```
sgratiy: have you checked whether this is a nwb2 file? We should probably test only nwb2 files.

tests/test_labnotebook.py
```diff
@@ -0,0 +1,40 @@
+from __future__ import print_function
```
sgratiy: we do not need `__future__` imports any longer

tests/test_labnotebook.py
gnayuy: Good point, have we uploaded any nwb2 file into tests/data folder? If not, I will upload one and update the test.

tests/test_labnotebook.py
gnayuy: Removed it.

tests/test_labnotebook.py
gnayuy: uploaded a NWB file 2.2.0 "Ctgf-T2A-dgCre;Ai14-495723.05.02.01-compressed-V2.nwb" to the tests/data folder

PR #359, opened 2020-03-12
[WIP] 286/integrate

sgratiy:
This branch integrates work from the team and includes all the refactored classes. 
Some interfaces may be a bit off and the tests are missing for nwb data readers.


ipfx/dataset/ephys_data_inteface.py
```diff
@@ -266,8 +267,13 @@ def get_acquisition_groups(self) -> List[str]:
 
         raise NotImplementedError
 
+<<<<<<< HEAD
```
NileGraddis: uh, prob should not be here

Issue #366, opened 2020-03-17
Fully typed and documented `EphysDataSet`


NileGraddis:
This is part of #361. We are refactoring `EphysDataSet` ([see here](https://github.com/AllenInstitute/ipfx/blob/286/refactor-nwb-reader/ipfx/dataset/ephys_data_set.py)). This issue is to document and type methods & attributes.

Issue #361, opened 2020-03-13
define new EphysDataSet interface in detail


NileGraddis:
We have a [new, source-independent interface](https://github.com/AllenInstitute/ipfx/blob/1f2a56ac02e18ee7cc1c1fd0d7990e1491e4e15a/ipfx/dataset/ephys_data_set.py#L10) for accessing ephys data, but a couple of questions remain ([background](https://github.com/AllenInstitute/ipfx/issues/355#issuecomment-598403788)). In particular, we should answer:
1. for each method defined on `EphysDataSet` if the return type does not fully describe the returned data (e.g. because it is a dataframe), we must document the returned data in more detail
2. for each method defined on `EphysDataSet` we must decide if the method should be part of our public interface.
3. for each attribute which makes up part of the public interface of `EphysDataSet`, we must ensure that this attribute is clearly defined, rather than implicitly set in a method.
4. For methods and attributes of the old `EphysDataSet` but not defined on the new one, we need to have a clearly state if, how, and when this functionality will be supported.

EDIT: I've done some investigation (see comment). Here is what I think remains to be done for this issue:
- [x] make sure `dataset.EphysDataSet` has all required attributes at the correct level of privacy (#366)
- [x] make sure each such attribute has typing (#366)
- [x]  make sure each such attribute is documented (#366)
- [x] make sure `EphysDataSet` methods are not doing `EphysDataInterface`'s work (#366)
- [x] tests for `EphysDataSet` (weirdly, an allensdk issue: https://github.com/alleninstitute/allensdk/issues/1433)

NileGraddis:
How is EphysDataSet constructed and used?

Construction:
=============

not constructed
---------------
`EphysDataSet` is frequently used as a class object, particularly to access its constants.

- **data_set_features**
    - `EphysDataSet.SHORT_SQUARE`
    - `EphysDataSet.RAMP`
    - `EphysDataSet.LONG_SQUARE`
- **sweep_props**
    - `EphysDataSet.SWEEP_NUMBER`
    - `EphysDataSet.STIMULUS_UNITS`
    - `EphysDataSet.STIMULUS_AMPLITUDE`
    - `EphysDataSet.STIMULUS_CODE`
    - `EphysDataSet.STIMULUS_NAME`

by helper function
------------------

Old version (`data_set_utils`) defines:
```Python
def create_data_set(
    sweep_info=None, 
    nwb_file=None, 
    ontology=None, 
    api_sweeps=True, 
[...]

Issue #353, opened 2020-03-11
nwb_reader integrate components


wbwakeman:
Follow-up to nwb_reader issue, need to integrate all individual components
ensure tests run successfully (resolve any issues that may arise)

NileGraddis:
I'm having to do this as part of #372. If someone has ideas for integration work that won't overlap (see draft pr #374 for my current state) feel free to assign yourself and get going.

PR #370, opened 2020-03-19
docs, types, implementations for ephys data set

NileGraddis:
This PR will cover #361 as well as part of #353. It already covers #366.

I'm leaving it as a draft until the tests are done (so full review is not expected), but would appreciate you guys looking at the new `EphysDataSet` interfaces now. Also @gnayuy  and @sgratiy please take a look at the changes to `EphysDataInterface` which impact (hopefully simplify!) your testing work for the derived interfaces.


NileGraddis:
The circle build is failing because a labnotebook reader test cannot download something :/ I don't think it is relevant here. Might need to check in with Thomas about the status of those tests.

NileGraddis:
@gnayuy I changed that back. `ipfx.stimulus` is part of our public API - we should not move it (breaking all those imports) without a good reason.

ipfx/dataset/ephys_data_interface.py
```diff
@@ -256,19 +163,7 @@ def get_sweep_map(self, sweep_number):
...
-        Returns
-        -------
-        names of acquisition groups
-        """
-
-        raise NotImplementedError
-
-    @abc.abstractmethod
-    def get_stimulus_groups(self) ->List[str]:
+    def get_stimulus_groups(self) -> List[str]:
```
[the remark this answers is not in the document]
sgratiy: Also, if we end up keeping it, the docstring incorrectly refers to the acquisition groups

ipfx/dataset/ephys_data_set.py
```diff
@@ -1,10 +1,23 @@
-from typing import Optional, List, Dict, Any, Tuple
-import warnings
+from typing import (
+    Optional, List, Dict, Tuple, Collection, Sequence, Union
```
sgratiy: what is the benefit of wrapping imports into parentheses?

ipfx/dataset/ephys_data_set.py
```diff
@@ -33,57 +46,79 @@ class EphysDataSet(object):
...
-                 sweep_info: Optional[Dict[str, Any]] = None):
-
-        self.data = data
-        self.ontology = data.ontology
-
-        if sweep_info is None:
-            sweep_info = self.extract_sweep_info()
+    @property
+    def ontology(self) -> StimulusOntology:
+        """The stimulus ontology maps short-form stimulus codes to complete
```
sgratiy: Not exactly: it maps codified description of the stimulus type to the meaningful description.  The stimulus code may actually be longer than stimulus name

ipfx/dataset/ephys_data_set.py
```diff
@@ -33,57 +46,79 @@ class EphysDataSet(object):
...
+        self._data: EphysDataInterface = data
 
-        sweep_info = []
-
-        for index, sweep_map in self.data.sweep_map_table.iterrows():
-            sweep_num = sweep_map["sweep_number"]
-            sweep_info.append(self.data.extract_sweep_record(sweep_num))
+    def _setup_stimulus_repeat_lookup(self):
+        """Construct a mapping from sweep number to the number of presentations 
+        of the stimulus presented during that sweep before and during that
```
sgratiy: this is a confusing docstring "presented during that sweep before and during that..."

ipfx/dataset/ephys_data_set.py
```diff
@@ -92,47 +127,75 @@ def filtered_sweep_table(self,
 
         if stimuli:
             mask = st[self.STIMULUS_CODE].apply(
-                self.ontology.stimulus_has_any_tags, args=(stimuli,), tag_type="code")
```
sgratiy: Are you sure you want to touch this? Are you confident that this will not break something?

ipfx/dataset/ephys_data_interface.py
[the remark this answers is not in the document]
NileGraddis: let's kill it

ipfx/dataset/ephys_data_set.py
NileGraddis: multiline, slightly prettier than \

ipfx/dataset/ephys_data_set.py
NileGraddis: nope, this is a mistake (I was printing here and deleted too much).

ipfx/dataset/ephys_data_interface.py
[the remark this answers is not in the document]
NileGraddis: hmm, the sweep map stuff depends on this, leaving it in the nwbdata

ipfx/dataset/ephys_data_set.py
NileGraddis: oh wait I thought this was in stimulus. This is fine, I just inserted line breaks.

ipfx/dataset/ephys_data_interface.py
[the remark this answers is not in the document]
sgratiy: Indeed, but sweep_map also depends on the `get_acquisition_groups` which must also stay then

Issue #345, opened 2020-03-05
use a proper authentication system for lims queries


NileGraddis:
We should not have [this kind of stuff](https://github.com/AllenInstitute/ipfx/blob/74904808979dab54125df23de0d0e27aceba9207/ipfx/lims_queries.py#L6) committed. Instead, we should follow the lead of [allensdk](https://github.com/AllenInstitute/AllenSDK/blob/cc6e3870af87606ed562aa7808ce667fe6efafe5/allensdk/core/authentication.py#L1) (by "follow the lead of", I mean "import and use").

NileGraddis:
I had to do this as part of my work on #372

PR #374, opened 2020-03-20
372/remove old dataset

NileGraddis:
This makes a large number of changes aimed at the 1.0.0 release (this pr brings the version to 1.0.0.a1). These changes broadly address
- removal of support for NWB1
    - this also changes the interface by which `EphysDataSet` is constructed, since an external h5 file is no longer required
- replacement of the old `EphysDataSet` code (this addresses #353)
- removal of python2 support shims
- updated credential handling (remove hardcoded passwords; this addresses #345)
- other misc cleanup (removing dead code, unused imports, the like)

I am currently addressing:
- [x] replace feature vector nwb-dependent tests with in-memory unit tests (see pr #378 for this work)
- [ ] use e.g. git lfs to store NWB2 data for examples (currently these query warehouse) EDIT: moved this to issue #379, since this pr is already enormous

I have created new issues  to handle some large tasks that this turned up:
- #377 fixing the run_feature_vector integration tests

This is a draft, but it is also huge. Please do take a look and jump in to fix any issues that you notice.


ipfx/stim_features.py
```diff
@@ -18,6 +17,7 @@ def get_stim_characteristics(i, t, test_pulse=True):
     start_idx = di_idx[start_idx_idx] + 1   # shift by one to compensate for diff()
     end_idx = di_idx[-1]
     if start_idx >= end_idx: # sweep has been cut off before stimulus end
+        print("foo", start_idx, end_idx, di_idx)
```
MatthewAitken: should this be more informative than just foo?

NileGraddis: that is a debug print, good catch :cactus:

PR #378, opened 2020-03-24
unit tests for feature vector

NileGraddis:
Some of the tests for the feature vector code are written as integration tests when they ought to be unit tests - they load up data from an nwb file and then call multiple functions in sequence. Much of this data is irrelevant to the functions being called, and the dependence on an nwb file makes these tests hard to run outside the institute (and hard to update for the move to nwb2). This pr replaces these tests with focused unit tests.

### DIFF
--- a/tests/test_feature_vector.py
+++ b/tests/test_feature_vector.py
@@ -14,14 +14,24 @@ import os
 from .helpers_for_tests import download_file
 
 
-TEST_OUTPUT_DIR = "/allen/aibs/informatics/module_test_data/ipfx/test_feature_vector"
+TEST_OUTPUT_DIR = os.path.join(
+    "/",
+    "allen",
+    "aibs",
+    "informatics",
+    "module_test_data",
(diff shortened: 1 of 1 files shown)


tests/test_feature_vector.py
```diff
@@ -725,51 +1184,23 @@ def test_isi_shape_skip_short():
...
     n_points = 100
     isi_norm = fv.isi_shape(test_sweep, test_spike_info, end, n_points=n_points)
     assert len(isi_norm) == n_points
 
     # Should only use second ISI
-    assert isi_norm[0] == (test_sweep.v[test_fast_trough_index[1]:test_fast_trough_index[1] + test_subsample].mean()
-        - test_threshold_v[1])
-
-
-def test_isi_shape_one_spike():
```
MatthewAitken: What happened to this one?

NileGraddis: that is ... a formatting error.

PR #383, opened 2020-03-26
BF(BK): do not pin hdmf and pynwb, only provide minimal compatible versions

yarikoptic:
This is a similar request to
https://github.com/AllenInstitute/AllenSDK/pull/1326
and please see
https://github.com/AllenInstitute/AllenSDK/issues/1325
for rationale, and pointed there
https://github.com/NeurodataWithoutBorders/pynwb/issues/1133#issuecomment-571764474

I consider it a "bug fix" (BF) since such hard pining (at the install_requires
level) prevents co-installation of this module and any other module requiring
different version of pynwb.  It would actually be ok to pin in requirements.txt but setup.py then must be adjusted to convert `==` into `>=` or alike. ATM it just passes them as is. See e.g. what pynwb does now: https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/setup.py#L15

I also marked it as BK since most likely it would break tests.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,9 +8,9 @@ pillow
 argschema
 allensdk
 dictdiffer
-hdmf==1.1.2
+hdmf>=1.1.2
 pyabf<2.3.0
-pynwb==1.0.2
+pynwb>=1.0.2
 six
 watchdog
 pg8000


yarikoptic:
I didn't even have to "go far" to experience the problem due to pinning: cannot even `pip install` current state of ipfx satisfying similar strict requirements of allensdk:
```shell
(git)lena:~/proj/dandi/nwb/ipfx[master]git
$> py=3; d=venvs/dev$py; virtualenv --python=python$py  $d && source $d/bin/activate && pip install -e .        
Already using interpreter /usr/bin/python3
Using base prefix '/usr'
New python executable in /home/yoh/proj/dandi/nwb/ipfx/venvs/dev3/bin/python3
...
ERROR: allensdk 1.6.0 has requirement argschema<2.0.0, but you'll have argschema 2.0.1 which is incompatible.
ERROR: allensdk 1.6.0 has requirement hdmf==1.0.2, but you'll have hdmf 1.1.2 which is incompatible.
ERROR: allensdk 1.6.0 has requirement pandas<=0.25.3,>=0.25.1, but you'll have pandas 1.0.3 which is incompatible.
...
```
pip does exit without error, so pretty much ignores the pinning being satisfied that it got at least some version of those installed.  Now, that allensdk possibly never was tested with those newer versions you require (since they pinned as well), things could break quite easily if work at all.

wbwakeman:
ipfx has been updated to use unpinned pynwb and hdmf.  This PR can be closed.

yarikoptic:
Great, thanks!

PR #387, opened 2020-03-30
fix gallery (new pr for debugging rtd statuses)

NileGraddis:
https://ipfx--387.org.readthedocs.build/en/387/auto_examples/index.html

- gets the gallery of examples running again
- ipfx auto-builds on pull requests (for some reason, it isn't posting statuses back. This feature is working for allensdk and neuron_morphology, which were updated at the same time. I've reached out to readthedocs' support)
- configures readthedocs in a yaml (vs the gui)
- calculate_dvdt removes duplicate time samples, not nans in general
- EphysDataSet trims trailing zeros, rather than NANing them. 
- version stored in single location (& bumped to 0.1.1)

@gouwens I would particularly appreciate your scientific review of the gallery examples. Do they look OK?

EDIT: readthedocs has just merged a bugfix that should fix the status posting.


docs/gallery/analysis_examples/ramp_analysis.py
```diff
@@ -24,13 +25,15 @@
...
 # Build the data set and find the ramp sweeps
-data_set = AibsDataSet(sweep_info=sweep_info, nwb_file=nwb_file)
-ramp_table = data_set.filtered_sweep_table(stimuli=data_set.ontology.ramp_names)
+data_set = create_data_set(sweep_info=sweep_info, nwb_file=nwb_file)
+ramp_table = data_set.filtered_sweep_table(
+    stimuli=data_set.ontology.ramp_names
+)
 ramp_sweep_set = data_set.sweep_set(ramp_table.sweep_number)
 
 # Build the extractors (we know stimulus starts at 0.27 s)
```
gouwens: This is probably not true for the ramp stimulus either? Seems likely copied from the long-square example.It works because spikes don't happen until long after this time and the analysis doesn't do anything with the stimulus waveform. But things like latency would not be accurately calculated.

docs/gallery/analysis_examples/short_square_analysis.py
```diff
@@ -23,17 +25,22 @@
...
+    stimuli=data_set.ontology.short_square_names
+)
 shsq_sweep_set = data_set.sweep_set(shsq_table.sweep_number)
 
-# Estimate the dv cutoff and threshold fraction (we know stimulus starts at 0.27s)
-dv_cutoff, thresh_frac = ft.estimate_adjusted_detection_parameters(shsq_sweep_set.v,
-                                                                   shsq_sweep_set.t,
-                                                                   0.27, 0.271)
+# Estimate the dv cutoff and threshold fraction 
+# (we know stimulus starts at 0.27s)
```
gouwens: Again, this is probably the wrong time for stimulus start?

docs/gallery/analysis_examples/short_square_analysis.py
```diff
@@ -23,17 +25,22 @@
...
+dv_cutoff, thresh_frac = estimate_adjusted_detection_parameters(
+    shsq_sweep_set.v, shsq_sweep_set.t, 0.27, 0.271
```
gouwens: These should be adjusted so that the detection parameters are actually estimated correctly. It's supposed to be right at the start of the (usually large) step current stimulus, because that rapid change in voltage can sometimes look like the start of a spike (and we don't want a spurious detection there).

ipfx/stimulus.py
```diff
@@ -26,7 +26,10 @@ def has_tag(self, tag, tag_type=None):
 
 class StimulusOntology(object):
 
-    DEFAULT_STIMULUS_ONTOLOGY_FILE = os.path.join(os.path.dirname(__file__), 'defaults/stimulus_ontology.json')
+    DEFAULT_STIMULUS_ONTOLOGY_FILE = os.path.join(
+        os.path.dirname(__file__), 
+        'defaults/stimulus_ontology.json'
```
gouwens: Should this be `"defaults", "stimulus_ontology.json"` to remove the slash from a path name? I don't actually know how `os.path.join` handles that cross-platform.

NileGraddis:
OK - I've gone through the analysis examples and used the stimulus epoch in place of the hardcoded start times

PR #391, opened 2020-04-01
rc/1.0.0

NileGraddis:
This is a release candidate for 1.0.0. New features should target this branch.


sgratiy:
We will use the GitFlow workflow that calls for creating a `release` branch that is then merged into master.

Issue #409, opened 2020-05-12
Update contribution guidelines, add templates


wbwakeman:
To the extent possible, use what has been done for AllenSDK.
Anything else needed to release ipfx
Release plan
instructions
Change log page? can use tag annotations on 'Releases' page

PR #414, opened 2020-05-27
Contribution guidelines

sgratiy:
Relates to: #409


CHANGELOG.md
```diff
@@ -0,0 +1,13 @@
+# Change Log
```
NileGraddis, line 1: decided not to go with github releases?

sgratiy: No I did not. Should we not use CHANGELOG.md if we use github release? How does it work?

sgratiy:
@MatthewAitken could you please elaborate on you point
>little bit more info on how that works to the release process

CHANGELOG.md
NileGraddis: well I guess we don't need a markdown changelog if we are going with github. Or we could track  "latest changes" in the repo while the full log is on the releases (as in allensdk).

sgratiy: I have updated the guidelines, stating that release notes are pulled from the changelog which in turn must be updated by committers.

Issue #360, opened 2020-03-12
Add more Multiclamp Commander Settings to .extract_sweep_stim_info() in AibsDataSet


kasbaker:
We need all the settings listed in .register_enabled_names(), which is a method of LabNotebookReader. Additionally, there should be settings for Cp Fast and Cp Slow, which each have a capacitance and time constant associated with them. I could not find these settings in .register_enabled_names().

tmchartrand:
Just wanted to echo that this is an important feature and it would be great to get @kasbaker 's fixes merged soon. 
While the get_notebook_value solution gets part of the way there, I would suggest that at least those values relevant to current clamp sweeps (namely capacitance neutralization) be added to get_sweep_metadata by default as in the original PR. This would allow the features to be accessible from the sweep table and in the output json, to simplify querying across cells/sweeps without actually loading the NWB. I can file a separate PR for this, since I've already been running it on my fork.
Potentially, the longer list of voltage-clamp metadata could be added also, conditional on a check of the sweep's clamp type, which would minimize any performance hit from repeatedly checking for properties that are not present.

PR #418, opened 2020-05-29
Issue #360 more MCC settings

kasbaker:
Added more settings from multiclamp commander to the sweep table. Relates to https://github.com/AllenInstitute/ipfx/issues/360


NileGraddis:
Hi @kasbaker. Thank you for the PR!

These changes seem reasonable, but I have a few questions:
1. IIRC these fields are not required on a PatchClampSeries subclass. If we run this code on an NWB file which has not written one of these fields, do we get a sweep table?
2. Where can a user look to find these values' definitions? Is this place easy to find for someone who is not poking around in our code?

kasbaker:
Putting this pull request on hold for now because `get_sweep_metadata()` is slow and this change will just make it slower.

Issue #444, opened 2020-06-11
don't try to add spikes if they already exist


wbwakeman:
If spikes already exist, use them.  Don't throw ValueError: 'spikes' already exists in 'root' error.

Also, add a flag so that when running this in the Institute  pipeline to indicate that we want to add the spikes

PR #446, opened 2020-06-11
Gh 444/use spikes if already exist

sgratiy:
# Addresses:
Add an optional write_spikes with the default in the run_feature_extractor set to True
but in the run_from_nwb_file the default is set to False
this would avoid making changes to LIMS. Users processing dandi files would by default have write_spikes=False.

Addresses issue [#444]

# Type of Fix:

# Solution:
Do not add spikes processing module if it already exists

### DIFF
--- a/ipfx/nwb_append.py
+++ b/ipfx/nwb_append.py
@@ -47,19 +47,26 @@ def append_spike_times(input_nwb_path: PathLike,
     nwb_io = pynwb.NWBHDF5IO(nwb_path, mode='a', load_namespaces=True)
     nwbfile = nwb_io.read()
 
-    # Create spike module
-    spike_module = ProcessingModule(name='spikes',
-                                    description='detected spikes')
-    for sweep_num, spike_times in sweep_spike_times.items():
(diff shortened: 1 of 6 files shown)


MatthewAitken:
Does this need a Lims update to write spikes in the job now?

ipfx/nwb_append.py
```diff
@@ -47,19 +47,26 @@ def append_spike_times(input_nwb_path: PathLike,
...
+                            unit='seconds',
+                            data=wrapped_spike_times,
+                            name=f"Sweep_{sweep_num}")
+            spike_module.add_data_interface(ts)
+
+        nwbfile.add_processing_module(spike_module)
+
+        nwb_io.write(nwbfile)
+    else:
+        raise ValueError("Cannot add spikes times to the nwb file: "
```
NileGraddis, line 68: do you need to raise a new valueerror here?

NileGraddis: vs. just reraise (maybe with some added info)

sgratiy:
@MatthewAitken I have changed the logic to avoid changes in LIMS. Thanks

Issue #454, opened 2020-06-25
qc_figs_dir should be relative to the cell output directory


sgratiy:
**Describe the bug**
qc_figs_dir should be always included inside the cell output dir rather then being an arbitrary output path because otherwise user would not be able to relate figure generated to the cell output folder. Also, this makes a more organized output and avoids a need to create an empty qc_figs_dir. The input.json should include the path to the qc_figs_dir

**Desired behavior**
instead of `qc_fig_dir`  add `plot_figures` parameter with the default True. It will create a directory in an expected location inside the cell folder and add it to the input.json.

NileGraddis:
@sgratiy [LIMS argues](http://stash.corp.alleninstitute.org/projects/TECH/repos/lims/browse/app/strategies/ephys_feature_extraction_v3_strategy.rb#55,18,20) qc_fig_dir as a full path to run_feature_extraction. The LIMS change to stop doing this would be pretty minor, though other users might be broken by the removal of qc_fig_dir. Maybe we can avoid breaking changes entirely though. Why not:
1. make qc_fig_dir default to a subdirectory of the outputs
2. add a plot_qc_figures parameter that controls whether figures are generated.

The upside: no breaking changes, and we support (probably rare) users who really want to organize their outputs differently. The downside: a bit more complication for a pretty niche case (that happens to be implemented in LIMS). What do you think @sgratiy @wbwakeman ?

sgratiy:
@NileGraddis, I agree on avoiding breaking changes in LIMS - this is how I envisioned when describing the issue. I have not argued to remove qc_fig_dir from the input.json
In fact, I would not let users to specify an arbitrary location for the qc_fig_dir.
I would use an optional `plot_figures` command line argument instead that would construct the path.

NileGraddis:
I don't think I understand your proposed solution. Are you saying that we still have qc_fig_dir, but reject paths that are not inside the output directory or ignore it entirely?

sgratiy:
Users should have a command line argument like `plot_qc_figures` for the run_pipeline_from_nwb_file. If `plot_qc_figures`=True, then qc_fig_dir is constructed to a subdir of the outputs. If `plot_qc_figures`=False, then qc_fig_dir=None. Makes sense?

NileGraddis:
What happens if a user provides both qc_fig_dir and plot_qc_figures? Does the behavior depend on where the specified qc_fig_dir is? Can a user even supply their own_qc_fig_dir at all?

PR #456, opened 2020-06-26
GH-454 run_pipeline_from_nwb_file is relative to cell output directory

NileGraddis:
- run_pipeline_from_nwb_file's qc_fig_dir argument is now relative to cell_dir (can still specify an absolute path if you like)
- can argue --qc_fig_dir with build figures and store them in "qc_figs"
- adds documentation for run_pipeline_from_nwb_file's parameters

I had to choose between consistency with other executables and consistency with the other parameters of this executable. I went with the latter, but I'm not particularly wedded to the choice. Let me know in your review what you think

### DIFF
--- a/ipfx/bin/run_pipeline_from_nwb_file.py
+++ b/ipfx/bin/run_pipeline_from_nwb_file.py
@@ -18,16 +18,41 @@ def main():
     parser = argparse.ArgumentParser(
         description="Process an nwb file through the ephys pipeline"
     )
-    parser.add_argument("input_nwb_file", type=str)
-    parser.add_argument("output_dir", type=str)
-    parser.add_argument("--write_spikes",
-                        type=bool,
-                        default=False,
-                        help="If true will attempt to append spike times to the nwb file",
-                        )
-    parser.add_argument("--input_json", type=str, default="input_json")
-    parser.add_argument("--output_json", type=str, default="output_json")
-    parser.add_argument("--qc_fig_dir", type=str, default=None)
+    parser.add_argument(
+        "input_nwb_file", type=str, help="process this NWB2 file"
+    )
+    parser.add_argument(
+        "output_dir", type=str, help="outputs will be written here"
+    )
+    parser.add_argument(
+        "--write_spikes", type=bool, default=False,
+        help="If true will attempt to append spike times to the nwb file",
+    )
+    parser.add_argument(
+        "--input_json", type=str, default="input_json",
+        help=(
+            "write pipeline input json file here (relative to "
+            "OUTPUT_DIR/cell_name, where cell_name is the extensionless "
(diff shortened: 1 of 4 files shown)


ipfx/bin/run_pipeline_from_nwb_file.py
```diff
@@ -18,16 +18,41 @@ def main():
...
+            "file)"
+        )
+    )
+    parser.add_argument(
+        "--qc_fig_dir", type=str, default=None, const="qc_figs", nargs="?",
+        help=(
+            "Generate qc figures and store them here (relative to "
+            "OUTPUT_DIR/cell_name, where cell_name is the extensionless "
+            "basename of the input nwb file). If you supply --qc_fig_dir with " 
+            "no arguments, the path will be OUTPUT_DIR/cell_name/qc_figs"
```
sgratiy, line 53: It is worth mentioning that the default value will make no figures.

ipfx/bin/run_pipeline_from_nwb_file.py
```diff
@@ -18,16 +18,41 @@ def main():
...
+    parser.add_argument(
+        "--output_json", type=str, default="output_json", 
+        help=(
+            "write output json file here (relative to OUTPUT_DIR/cell_name, "
+            "where cell_name is the extensionless basename of the input NWB "
+            "file)"
+        )
+    )
+    parser.add_argument(
+        "--qc_fig_dir", type=str, default=None, const="qc_figs", nargs="?",
```
sgratiy, line 48: Nice!

ipfx/bin/run_pipeline_from_nwb_file.py
```diff
@@ -18,16 +18,41 @@ def main():
...
+    )
+    parser.add_argument(
+        "output_dir", type=str, help="outputs will be written here"
+    )
+    parser.add_argument(
+        "--write_spikes", type=bool, default=False,
+        help="If true will attempt to append spike times to the nwb file",
+    )
+    parser.add_argument(
+        "--input_json", type=str, default="input_json",
```
sgratiy, line 32: it would be better to name it `input.json`- i.e., with an extension. This allows configuring preferred editor for opening files based on extension.

sgratiy: Also, check that quick_start page refers to the correct name of the input/output JSON. The last time I saw, it was referring to  `pipeline_input.json` and `pipeline_output.json`

NileGraddis: good catch, thank you!

ipfx/bin/run_pipeline_from_nwb_file.py
NileGraddis: done

Issue #473, opened 2020-09-04
unequal number of datapoints in NWB conversion


ozsat:
I noticed that when I convert DAT to NWB (`ipfx.bin.run_x_to_nwb_conversion.convert`), the voltage samples and current samples are not equal. In the current samples, 5 of the given points are always missing from the beginning. It seems to be a major problem because the ipfx cant process those NWB files. 
I checked the raw data and it contains an equal number of points (I also attached a picture that shows the discrepancy in the NWB file). 

![voltage_samples_nwb](https://user-images.githubusercontent.com/36845530/92212711-afc85880-ee92-11ea-8e56-1a6e36e7b135.png)
![stimulus_samples_nwb](https://user-images.githubusercontent.com/36845530/92212714-b0f98580-ee92-11ea-9d4e-f06486ac8e53.png)

When I run `run_pipeline` on the generated NWB data the following error message appears:

INFO:root:
INFO:root:***** ***** ***** Extract QC features ***** ***** *****
INFO:root:=======================================================
INFO:root:
INFO:root:Ontology is not provided, using default E:\anaconda3\lib\site-packages\ipfx\defaults\stimulus_ontology.json
E:\anaconda3\lib\site-packages\ipfx\qc_feature_extractor.py:28: VisibleDeprecationWarning: Function get_sweep_number is deprecated. call .get_sweep_numbers()[-1] instead
  blowout_sweep_number = data_set.get_sweep_number(ontology.blowout_names)
E:\anaconda3\lib\site-packages\ipfx\stimulus.py:88: UserWarning: Could not find stimulus: extpinbath
  warnings.warn("Could not find stimulus: %s" % tag)
E:\anaconda3\lib\site-packages\ipfx\stimulus.py:88: UserWarning: Could not find stimulus: extpbreakn
  warnings.warn("Could not find stimulus: %s" % tag)
E:\anaconda3\lib\site-packages\ipfx\stimulus.py:88: UserWarning: Could not find stimulus: Long square
  warnings.warn("Could not find stimulus: %s" % tag)
E:\anaconda3\lib\site-packages\ipfx\stimulus.py:88: UserWarning: Could not find stimulus: Rheobase
  warnings.warn("Could not find stimulus: %s" % tag)
E:\anaconda3\lib\site-packages\ipfx\stimulus.py:88: UserWarning: Could not find stimulus: Short square
  warnings.warn("Could not find stimulus: %s" % tag)
[...]

PR #474, opened 2020-09-09
GH-471 Update pynwb requirements

sgratiy:
Addresses #471
- [x] Update version range of pynwb, drop hdmf requirement since it is installed as a dependency of pynwb
- [x] Build dedicated image for ipfx. 
- [x] Updated bamboo task to use the dedicated image both for Github and nightly build plans

Notes: 
1. I hosted in on the Docker Hub since I am not sure how to get the image on the aibs-artifactory.corp.alleninstitute.org. When Nick is back we can ask him how to upload the image to the artifactory.
2. This PR is against mastersince this is a hotfix. May be not so hot, but a fix anyway :)

Validation: 
the github branch CI on bamboo is passing!

PR #475, opened 2020-09-14
fix segment length rounding error #473 

tmchartrand:
Addresses #473 
In the DAT converter, segments of sweeps are constructed with length calculated by floating point division. The result can be slightly less or greater than the the actual integer value, so applying math.trunc was dropping data points in some cases. This fixes the issue by rounding instead, and raises an error if the unrounded value is not close to an integer.

### DIFF
--- a/ipfx/x_to_nwb/hr_segments.py
+++ b/ipfx/x_to_nwb/hr_segments.py
@@ -159,7 +159,11 @@ class Segment(ABC):
         """
         Return the number of points of this segment.
         """
-        return math.trunc(duration / self.sampleInterval)
+        num_points = duration / self.sampleInterval
+        num_points_int = int(np.round(num_points))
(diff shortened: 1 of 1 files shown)


tmchartrand:
These failures seem to be related to pynwb/hdmf versions, not this PR - is that being addressed?

wbwakeman:
Yes, a new release should be available soon
#474

wbwakeman:
@tmchartrand The pynwb/hdmf version issue was fixed with the 1.0.1 version released on September 16th.  Based on your comment, is this PR still needed?

tmchartrand:
@wbwakeman  yes, definitely still needed. That previous comment was referring to the tests that were initially showing as failing for this PR, not the underlying issue, sorry it was unclear.

sgratiy (approved):

Issue #469, opened 2020-08-19
Large number of cells missing long square features


tmchartrand:
I've discovered that quite a few recent cells have null records for long squares in their stored IPFX output (json output as queried from LIMS). I can't figure out why this is missing, since when I rerun them myself it succeeds, and no errors are recorded in the logs I can see.
An example
cell id: 839897427
output json: /allen/programs/celltypes/production/humancelltypes/prod259/Ephys_Roi_Result_839897299/EPHYS_FEATURE_EXTRACTION_V3_QUEUE_839897299_output.json
nwb file: /allen/programs/celltypes/production/humancelltypes/prod259/Ephys_Roi_Result_839897299/nwb2_H19.06.351.11.16.01.01.nwb

This might be a small issue, or it might be a big issue.
For that reason, we will timebox the investigation of this issue to 1/2 day. Any longer than that, then we need to plan how to address the issue, and schedule that in another sprint.

This appears to be related to the stimulus_ontology.json file that is generated by LIMS.  The next research segment needs to investigate why the LIMS-generated stimulus_ontology.json files result in null "Long Square" and "Cell" features.

tmchartrand:
To emphasize the magnitude of this issue, in a test set of ~200 recent inhibitory cells (newer unreleased data), about half seem to show this issue. For whatever reason it seems more common in this set than recently released excitatory patch-seq cells, but that may just be because they were re-processed more recently.

sgratiy:
@tmchartrand Could you please link a couple of ephys_roi_result from the recent inhibitory cell pool that show the issue?

tmchartrand:
@sgratiy  that info is in the issue - or is there something specific missing that you need?

sgratiy:
@tmchartrand I understood your comment about "test set of ~200 recent inhibitory cells" as if it was an additional set of data that exhibited this issue and though it would be good to have an additional example to work with.

tmchartrand:
I see, I can get you additional examples from that set at some point, just having VPN issues right now. I expect that example is representative of the failures though.

sgratiy:
It appears that LIMS generated stimulus_ontology is a likely reasons for this issue:
when ontology was generated in python we were getting:
[
      "code",
      **"EXTPGGAEND",**
      "EXTPGGAEND141203"
]
no with LIMS generated ontlogy we get this
    [
      "code",
      **"141203",**
      "EXTPGGAEND141203"
 ]
for all differences just
[...]

PR #481, opened 2020-10-02
Prevent mismatch between number of values and labels in AP feature plots 

sgratiy:
**Background:**
Sometimes AP feature value may be null. Previously these were treated as np.nan but at some point we changed np.nan to None that broke plotting.

This bug was discovered when validating http://jira.corp.alleninstitute.org/browse/PBS-2458 as a solution to #469 
The slow trough feature for the rheobase sweep is null for the nwb file in #469.

**Solution:**
allow None values into the plot which would silently omit the value from the plot as it did before the breaking change 
**Validation:**
should generating figures without errors:
`python -m ipfx.bin.run_pipeline_from_nwb_file /path/to/nwb2_H19.06.351.11.16.01.01.nwb /output/dir/ --qc_fig_dir
`

### DIFF
--- a/ipfx/plot_qc_figures.py
+++ b/ipfx/plot_qc_figures.py
@@ -84,18 +84,19 @@ def plot_single_ap_values(data_set, sweep_numbers, lims_features, sweep_features
 
     plt.figure(figs[0].number)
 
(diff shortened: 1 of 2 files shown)


wbwakeman (approved):
Cool.  With these changes I am able to run the LIMS feature extraction jobs that were previously failing

PR #491, opened 2021-01-05
GH-470 Add features state to the json output

### LINKED ISSUE #470: save feature_states in output
As far as I can see, the nice output from the record_errors decorator is not actually getting saved unless all features failed!
This would be very helpful for tracking down failures, as in #469, and should be a very simple fix.

sgratiy:
Resolves #470

Add features_state to the output and update tests

### DIFF
--- a/ipfx/bin/run_feature_extraction.py
+++ b/ipfx/bin/run_feature_extraction.py
@@ -63,7 +63,8 @@ def run_feature_extraction(input_nwb_file,
                         'sweep_features': sweep_features,
                         'cell_record': cell_record,
                         'sweep_records': sweep_records,
-                        'cell_state': cell_state
+                        'cell_state': cell_state,
+                        'feature_states': feature_states
                         }
 
         if write_spikes:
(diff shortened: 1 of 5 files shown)


MatthewAitken (approved):

isaak-willett:
Two questions,
1: Why do you access the last index on the sweeps? I'm not super familiar with the methods so this is more for my curiosity.
2: What's the validation criteria you used to ensure this PR is addressing the issue?

sgratiy:
@isaak-willett 

> 1: Why do you access the last index on the sweeps? I'm not super familiar with the methods so this is more for my curiosity.

The latest index is used because there should be only a single sweep of this kind (e.g. blowout)

> What's the validation criteria you used to ensure this PR is addressing the issue?

Sorry I have not made this explicit, now I have updated the PR with the added `feature_state` block in the output.json. Unfortunately output.json files are on git-lfs and do not allow to see the diff for the output.json.

isaak-willett (approved):

gnayuy (commented):
Looks good to me.

Issue #498, opened 2021-02-16
ipfx should handle sweeps with missing stimulus amplitudes


wbwakeman:
NWB2 files are being generated by the ephys rigs at AIBS now and being sent to LIMS. Most, or all, are failing because one of the sweeps has a null stimulus amplitude. Example shown below

Rusty says: 
> “this is something that does occur quite regularly as a byproduct of the real time QC functions that we are running during acquisition. If a sweep is started and the cell’s pre-stimulus baseline voltage is beyond +/- 1mV of the predetermined resting baseline, then the sweep will be terminated early, in many cases before the stimulus epoch is initiated, and so those sweeps will not have a stimulus amplitude associated with them.”

```
/allen/programs/celltypes/production/mousecelltypes/prod3000/Ephys_Roi_Result_1074702561/202101051415_EPHYS_NWB_STIMULUS_SUMMARY_V3_QUEUE_1074702561_output.json
 
{
      "sweep_number": 5,
      "stimulus_units": "Amps",
      "bridge_balance_mohm": 10.98901081085205,
      "leak_pa": null,
      "stimulus_scale_factor": 10.0,
      "stimulus_code": "X1PS_SubThresh",
      "stimulus_code_ext": "X1PS_SubThresh[1]",
      "clamp_mode": "CurrentClamp",
      "stimulus_name": "Long Square",
      "tags": [
        "Recording stopped before completing the experiment epoch"
      ],
      "stimulus_start_time": 1.5e-05,
      "stimulus_amplitude": null,   <-------------------------------------------------------------------------------------------------
      "stimulus_duration": 3.59998,
      "stimulus_interval": null
    },
```

t-b:
What Rusty says makes totally sense. Instead of looking into the acquired AD data, you could also look into the stimulus sets to get the amplitude. This works will all MIES versions. Depending on the stimulus set you could also try out the new epoch information, (available from https://github.com/AllenInstitute/MIES/commit/6191bae16f89e632b1a0f807ef9a4559824a5e4a on) which is available in the labnotebook. It holds various properties of the DA wave as key value pairs.

sgratiy:
I assume this concerns only current clamp sweeps as in example above.
In this case we can address the issue by looking into

` tags = check_sweep_integrity(sweep, is_ramp)
`
in qc_feature_extractor.py#L280

This function is checking for the presence of all the define epochs.
For the null stimulus_amplitude we could make stimulus epoch computed by
[...]

PR #503, opened 2021-03-03
returns 0.0 for amplitude instead of nan

MatthewAitken:
Bugfix, address #498 

Runs were failing because nan stimulus amplitudes converted to null when written to json, and LIMS EphysStimulus requires a numeric amplitude.

### DIFF
--- a/ipfx/stim_features.py
+++ b/ipfx/stim_features.py
@@ -32,6 +32,9 @@ def get_stim_characteristics(i, t, test_pulse=True):
     else:
         amplitude = float(peak_low)
 
+    if np.isnan(amplitude):
+        amplitude = 0.0
+
     return start_time, duration, amplitude, start_idx, end_idx
 
 
--- a/ipfx/stim_features.py
+++ b/ipfx/stim_features.py
@@ -33,7 +33,7 @@ def get_stim_characteristics(i, t, test_pulse=True):
         amplitude = float(peak_low)
 
     if np.isnan(amplitude):
-        amplitude = 0.0
+        return None, None, 0.0, None, None
 
     return start_time, duration, amplitude, start_idx, end_idx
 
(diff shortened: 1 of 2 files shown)


tmchartrand:
Sweeps with a true amplitude of 0 are possible, although rare, so this seems potentially confusing. Is it not possible to change the LIMS schema to allow null? If not, simply dropping these sweeps before passing to LIMS might be a better solution.

MatthewAitken:
> Sweeps with a true amplitude of 0 are possible, although rare, so this seems potentially confusing. Is it not possible to change the LIMS schema to allow null? If not, simply dropping these sweeps before passing to LIMS might be a better solution.

We could change the lims schema instead. I changed it this way because in the same function, a few lines up, it sets amplitude to 0.0 for two other error cases. I can change it to exactly replicate those other cases, and return None, None, 0.0, None, None if that is less confusing. Dropping sweeps at this stage seems to be against the design, but I'm not 100% on that.

tmchartrand:
Interesting - you're right that one error case seems to be trying to do exactly this (# sweep has been cut off before stimulus end). The other (# if no stimulus is found) is only sort of an error case, since no stimulus is truly zero amplitude. I hadn't realized that zero was already used for missing here, in that case it's probably not worth changing the approach.

What I'm puzzled by though is why this new case has come up, rather than being caught by the existing error case. Was there a change in the NWB files such that they now have nans in the stimulus? Previously I thought missing data was zeroed out in the NWB, then truncated in EphysDataSet.get_sweep_data, causing this stimulus check to hit the first error case rather than calculate a nan amplitude.

ipfx/stim_features.py
```diff
@@ -32,6 +32,9 @@ def get_stim_characteristics(i, t, test_pulse=True):
     else:
         amplitude = float(peak_low)
 
+    if np.isnan(amplitude):
```
sgratiy, line 35: I believe it would be better to move the lines:
```
        stim_features = current_clamp_sweep_stim_features(sweep)
        sweep_features.update(stim_features)
```
inside `if not tags:` condition on line 286, then you would not need to add nan handling here.

sgratiy (approved):
Looks good, but see my comment.

ipfx/stim_features.py
MatthewAitken: this would't work because stim_features (in particular stimulus amplitude) are required outputs, qc is not

PR #505, opened 2021-03-09
upgrades pip so cryptography wheel gets installed

MatthewAitken:
Bugfix - circleci deploy_pypi job is failing due to no rust compiler found. Rust is needed to build cryptography from source, but updating pip and cryptography should grab a wheel instead, so no need for Rust.


sgratiy (approved):
glad you resolved it

Issue #509, opened 2021-03-23
Cells failing QC due to "electrode_0_pa missing value"


ru57y34nn:
**Describe the use case that is addressed by this feature.**
There are currently nearly 700 IVSCC pipeline experiments that are failing cell QC with the fail tag "electrode_0_pa missing value". This fail tag is referring to one of our initial voltage clamp QC sweeps that IPFX is unable to find (stimulus name: ‘EXTPINBATH’) because the incorrect sweep was being ran at this point (stimulus name: ‘EXTPSMOKET’). This is actually not that big of a deal because these two stimulus waves are actually the same wave, but with different names to make it easy to distinguish what point we are at during the experiment simply by stimulus name, which is how IPFX identifies these sweeps.

**Describe the solution you'd like**
If the ‘EXTPINBATH’ sweep is not found then it needs to look for the ‘EXTPSMOKET’ sweeps and do a quick resistance calculation to ensure that the rig user actually had the pipette in the bath and then measure the baseline electrode current and compare it to the qc values to ensure that it is in range. This can be accomplished by adding a new function to qc_feature_extractor.py that will look for the last 'EXTPSMOKET' sweep before the 'EXTPCllATT' sweep and if one exists, then it should check the resistance by calling 'measure_seal' from qc_features.py, and if the seal is less than 10 MOhms then call 'measure_electrode_0' from 'qc_features.py'. This new function can be called by 'extract_electrode_0' in the try, except block when the 'EXTPINBATH' sweep is not found.

**Describe alternatives you've considered**
The only other alternative solution would be to manually QC all of these cells, and we would like to avoid that if possible.

**Additional context**
This is a high priority issue because this issue is affecting several hundred IVSCC pipeline experiments, all of which are failing QC due to the issue

**Do you want to work on this issue?**
I am already working on a fix for this issue and will submit a pull request as soon as I am done testing.

wbwakeman:
Hi @ru57y34nn I think the Marmot team can fix this if you can help us figure out what is the set of affected data.  Can you please provide a list of experiments, or another way to identify the data that has this issue?  Thank you

ru57y34nn:
[...]

PR #510, opened 2021-03-26
Fix missing electrode 0 when in bath sweep was ran with smoketest sti…

ru57y34nn:
…mulus

This commit extends the 'extract_electrode_0' function and tries to
determine if the 'in-bath' sweep was ran with the 'EXTPSMOKET' stimulus
when the 'EXTPINBATH' stimulus is not found.

If 'EXTPINBATH' is not found then measure the seal of all 'EXTPSMOKET'
sweeps that occur before the first instance of 'EXTPCllATT' and only if
the resistance of the seal is below 10 MOhms then call
'measure_electrode_0'. Electrode 0 is not available if no 'EXTPSMOKET'
sweeps occurred before the first 'EXTPCllATT' sweep, the resistance is
above 10 MOhms, or if no 'EXTPSMOKET' or no 'EXTPCllATT' sweeps found.

Thank you for contributing to the IPFX, your work and time will help to
advance open science! 

# Overview:
Many of the recent IVSCC pipeline experiments are failing cell QC with fail tag
"electrode_0_pa missing value". This issue occurred because a software issue,
which has since been fixed, that was causing the in-bath sweep to be collected 
with the 'EXTPSMOKET' stimulus instead of the 'EXTPINBATH' stimulus. In fact,
these stimuli are the same wave, but with different names so that the stage 
of the experiment can be determined simply by the stimulus wave name. 
Most of these failing experiments actually contain the in bath sweep, but it 
was ran with the 'EXTPSMOKET' stimulus. IPFX needs to be able to look for 
in-bath sweeps collected with the 'EXTPSMOKET' stimulus when the 
'EXTPINBATH' stimulus is not found.

# Addresses:
Addresses issue [#509](https://github.com/AllenInstitute/ipfx/issues/509)

# Type of Fix:
- [x] New feature (non-breaking change which adds functionality)

# Solution:
When 'extract_electrode_0' does not find any 'EXTPINBATH' stimulus sweeps,
then it will look through any 'EXTPSMOKET' stimulus sweeps that occurr
before the first instance of the 'EXTPCllATT' sweep and check the resistance.
If the resistance is low, < 10 MOhms, then the pipette is determined to
have been in the bath and the 'measure_electrode_0' is called and the value
is returned. Electrode 0 is determined to be unavailable if no 'EXTPSMOKET'
sweeps occurred before the first 'EXTPCllATT' sweep, if no 'EXTPSMOKET'
sweeps have a resistance < 10 MOhms, or if no 'EXTPSMOKET', or no 
'EXTPCllATT' sweeps are present.

# Changes:
- Added self.smoketest_names to StimulusOntology to find 'EXTPSMOKET'
- Extended extract_electrode_0 function to try to get electrode 0 from 
   'EXTPSMOKET' stimulus sweeps when 'EXTPINBATH' is not found

# Validation:
I have tested this on the 600+ pipeline experiments that are affected by
this issue and have validated that it is only able to extract the electrode 0
from 'EXTPSMOKET' sweeps when the 'EXTPINBATH' is not found and 
if the resistance is <10 MOhms (the range of an open pipette in the bath).
I wrote some unit test that use experiments where the 'EXTPINBATH' is
missing and either:
- there is a 'EXTPSMOKET' sweep but the resistance is too high
- there are multiple 'EXTPSMOKET' sweeps, some occurring after
   'EXTPCllATT', but only one before the 'EXTPCllATT' sweep that has 
   a resistance < 10 MOhms 
- there are multiple 'EXTPSMOKET' sweeps, all occurring before the 
  'EXTPCllATT' sweep but only one with resistance < 10 MOhms

### Unit Tests:
A note about my unit tests: They are not hermetic as I needed data_sets
to test the extract_electrode_0 function so I am using nwb files from LIMS.
Please let me know if this is not acceptable and if there is an alternative
solution I could use.

# Notes:
Please let me know if there are any issues with my solution so that we
can address those ASAP. A lot of IVSCC pipeline data is being affected
by this issue and we are in need of a fix that will accomplish the 
functionality that I have added here. Thanks

### DIFF
--- a/ipfx/qc_feature_extractor.py
+++ b/ipfx/qc_feature_extractor.py
@@ -91,35 +91,34 @@ def extract_electrode_0(data_set, tags):
     -------
     e0: float
     """
-
-    max_pipette_resistance = 0.0075 # 7.5 MOhms
+   
     ontology = data_set.ontology
 
     try:
         bath_sweep_number = data_set.get_sweep_numbers(ontology.bath_names)[-1]
         bath_data = data_set.sweep(bath_sweep_number)
-        seal = qcf.measure_seal(bath_data.i, bath_data.sampling_rate)
         
-        if abs(seal) <= max_pipette_resistance:
-            e0 = qcf.measure_electrode_0(bath_data.i, bath_data.sampling_rate)
+        e0 = qcf.measure_electrode_0(bath_data.i, bath_data.sampling_rate)
 
     except IndexError as e:  # ontology.bath_names did not find any bath sweeps
         try:
+            print('tyring smoketest')
             # collect all smoke test sweeps and the first cell attached sweep
             smoke_test_sweeps = data_set.get_sweep_numbers(ontology.smoketest_names)
             cell_attached_sweep = data_set.get_sweep_numbers(ontology.seal_names)[0]
-            
+            print(smoke_test_sweeps, cell_attached_sweep)
             # find only smoketest sweeps that were ran before cell attached sweep
             inbath_candidates = filter_smoketests(smoke_test_sweeps, cell_attached_sweep)
-
+            print(inbath_candidates)
             # check pipette resistance of inbath_candidates sweeps and append to 
-            # e0_candidates if pipette resistance is less than 10 MOhms. 
+            # e0_candidates if pipette resistance is less than max_pipette_resistance. 
             e0_candidates = []
+            max_pipette_resistance = 7.5   # MOhms
             for sweep in inbath_candidates:
                 bath_data = data_set.sweep(sweep)
-                seal = qcf.measure_seal(bath_data.v, bath_data.i, bath_data.t)
-                # print(seal)
-                if abs(seal) <= max_pipette_resistance:
+                pipette_resistance = qcf.measure_input_resistance(bath_data.v, bath_data.i, bath_data.t)
+                print(pipette_resistance)
+                if abs(pipette_resistance) < max_pipette_resistance:
                     e0 = qcf.measure_electrode_0(bath_data.i, bath_data.sampling_rate)
                     e0_candidates.append(e0)
 
--- a/ipfx/qc_feature_extractor.py
+++ b/ipfx/qc_feature_extractor.py
@@ -92,13 +92,16 @@ def extract_electrode_0(data_set, tags):
     e0: float
     """
 
+    max_pipette_resistance = 0.0075 # 7.5 MOhms
     ontology = data_set.ontology
 
     try:
         bath_sweep_number = data_set.get_sweep_numbers(ontology.bath_names)[-1]
         bath_data = data_set.sweep(bath_sweep_number)
-
-        e0 = qcf.measure_electrode_0(bath_data.i, bath_data.sampling_rate)
+        seal = qcf.measure_seal(bath_data.i, bath_data.sampling_rate)
+        
+        if abs(seal) <= max_pipette_resistance:
+            e0 = qcf.measure_electrode_0(bath_data.i, bath_data.sampling_rate)
 
     except IndexError as e:  # ontology.bath_names did not find any bath sweeps
         try:
@@ -115,8 +118,8 @@ def extract_electrode_0(data_set, tags):
             for sweep in inbath_candidates:
                 bath_data = data_set.sweep(sweep)
                 seal = qcf.measure_seal(bath_data.v, bath_data.i, bath_data.t)
-                print(seal)
-                if abs(seal) < 0.01:
+                # print(seal)
+                if abs(seal) <= max_pipette_resistance:
                     e0 = qcf.measure_electrode_0(bath_data.i, bath_data.sampling_rate)
                     e0_candidates.append(e0)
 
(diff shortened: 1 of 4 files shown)


MatthewAitken:
Looks good, but needs two changes:
* Target dev instead of master (I'll create a new release for this functionality so it can be put into production)
* test data in LIMS is indeed bad. Please create new directory "ipfx/tests/data/<smokedata or something descriptive>" and place the nwb files there. Before committing the test data please remove as much data from the nwb files as possible (ideally keep the files under 10MB each)

When it's done I'll handle all the release work! Thanks!

MatthewAitken (approved):
Looks good to me! I asked Sergey to also peak at it, and then I'll merge it in and create a new IPFX release!

sgratiy:
@ru57y34nn Since this is an issue with the data I can see an alternative approach to correct the annotation of the nwb files and LIMS records? Would not that be a better approach?

ipfx/qc_feature_extractor.py
```diff
@@ -59,9 +100,35 @@ def extract_electrode_0(data_set, tags):
...
+            inbath_candidates = filter_smoketests(smoke_test_sweeps, cell_attached_sweep)
+
+            # check pipette resistance of inbath_candidates sweeps and append to 
+            # e0_candidates if pipette resistance is less than 10 MOhms. 
+            e0_candidates = []
+            for sweep in inbath_candidates:
+                bath_data = data_set.sweep(sweep)
+                seal = qcf.measure_seal(bath_data.v, bath_data.i, bath_data.t)
+                print(seal)
+                if abs(seal) < 0.01:
```
sgratiy, line 119: How do you determine this threshold? Can it be set in terms of the existing qc parameters? What would happen if we did not include this condition here?

sgratiy: Up to this point, we did not have ad-hoc thresholds in the code and I would prefer us keeping it that way. Thus if you can think of alternative to setting a threshold 0.01 that would be preferable.

sgratiy (commented):
I feel that since we are dealing with the data issue, we should fix data, not the code processing it.
But if that is not feasible, then we must at least document the change to the QC evaluation. IPFX does not have this detailed documentation because we rely on the published white paper Electrophysiology Overview. But if we start to diverge from it we must describe what we do. At least we should describe how we deviate from the main path.

ipfx/qc_feature_extractor.py
ru57y34nn: @sgratiy I will look into this and see if there is a way that we can correct the data rather than modifying IPFX. The threshold is there because the EXTPSMOKET sweep is usually run with the pipette out of the bath, but for a period of time it was also being run once the pipette was in the bath in place of the EXTPINBATH. We need a way to determine if the pipette was in the bath during the EXTPSMOKET before using it to measure the electrode_0 value, and a low resistance seal (10 MOhm, or 0.01 Ohm) is the only way I can see to do that. The resistance of an open pipette tip in the bath is usually around 3-8 MOhms and will never exceed 10 MOhms (you simply can't patch with a pipette that has a resistance higher than 10 MOhms), and calculating a seal from a smoke test where the pipette is out of the bath will result in a very high seal (usually several GOhms). This is the best way I could figure out how to find EXTPSMOKET sweeps where the pipette was in the bath rather than out of the bath. I will give an update here once I figure out if there is an alternative route we can go here.

ru57y34nn: @sgratiy I see that the white paper states that "Thick walled borosilicate glass (...) electrodes were manufactured (...) with a resistance of 3 to 7 MOhms". Would it be acceptable if I made this threshold a constant and made the comparison to that (instead of the arbitrary 0.01 Ohm threshold defined in the middle of the code)? 

Also, while I see where you are coming from about fixing the data rather than changing the code that processes it, I can see that the extract_electrode_0 function isn't entirely robust as it is, in that it doesn't check the resistance of the pipette to ensure that it is in the range mentioned in the white paper. I think that the resistance check I am doing needs to also be applied even when getting the sweep with the proper in-bath name, EXTPINBATH. I could add another function to check pipette resistance to qc_features.py and this could be called in extract_electrode_0 to ensure that the resistance of the pipette is in range before getting the baseline current (e0). As it is right now, any sweep that is ran with the EXTPINBATH stimulus will have the baseline current extracted even if the pipette was not in the bath, or in some other configuration, and if that baseline current is within range then electrode_0 will pass QC. 

The Ephys team's thoughts on this are that the data are immutable and that the purpose of ipfx is to make use of the data and that it should be able to handle certain errors in the data. I think that it is reasonable for ipfx to check through the smoketest sweeps, EXTPSMOKET, when no sweep with the in-bath name, EXTPINBATH, is found, and see if one of those sweeps was collected with the pipette in the bath, especially because these stimulus waves are exactly the same. The EXTPSMOKET and the EXTPINBATH stimulus waves are exactly the same, just named differently to more easily identify where in the experiment the user should be; however it should not be assumed the user always runs the appropriate stimulus.

I don't think that what I am proposing here deviates from the white paper at all but let me know what you think Sergey, thanks!

sgratiy:
Closing this PR without merging. Will correct data instead to solve the underlying issue.

Issue #508, opened 2021-03-10
Correctly infer stimulus/experiment epochs from traces with Nan values


wbwakeman:
Currently the "Recording stopped before completing the experiment epoch" tag is preventing sweeps from being added to the "sweep_states" list in EPHYS_QC_V3_QUEUE_output.json.  Need to update sweep_states with the tag.

Touch base with Rusty about his expectations for these tags and qc calls at the level of 'sweep' and 'cell'.

Example: /allen/programs/celltypes/production/mousecelltypes/prod3052/Ephys_Roi_Result_1086833106/202103081352_EPHYS_QC_V3_QUEUE_1086833106_output.json shows that the cell did not fail ephys qc (“failed_qc”: false), and all of the sweeps are now in LIMS and they are all “auto passed”, but in the ephys qc json there is only a single sweep that shows up in the “sweep_states” list:
``` 
  "sweep_states": [
    {
      "sweep_number": 54,
      "passed": true,
      "reasons": []
    }
  ]
```
 Looking closer at /allen/programs/celltypes/production/mousecelltypes/prod3052/Ephys_Roi_Result_1086833106/202102261501_EPHYS_NWB_STIMULUS_SUMMARY_V3_QUEUE_1086833106_output.json, sweep 54 is the only sweep without the tag “Recording stopped before completing the experiment epoch”. This tag should not prevent sweeps from passing ephys QC and then feature extraction, since most sweeps will have this tag because they auto-terminate the sweeps after the stimulus epoch and once the cell returns to baseline in order to save time during experiments.

Rusty checked some other experiments and confirms that only sweeps without “tags” in the EPHYS_NWB _STIMULUS_SUMMARY_V3_QUEUE_output.json are making it into the “sweep_states” list in the EPHYS_QC_V3_QUEUE_output.json.

Would like some clarity as to how all of the sweeps are showing up in LIMS as “auto passed” but they are omitted from the “sweep_states” of the EPHYS_QC_V3_QUEUE_ouput.json because of tags that are not necessarily indicating an issue.

tmchartrand:
Just a thought here that for “Recording stopped before completing the experiment epoch” in particular, I'm not sure just ignoring the tag is the answer. The QC checks currently assume the full epoch is present to calculate delta_vm. We may need to either drop that check (since it's essentially duplicating a MIES QC step), or tweak it to be more compatible with the MIES approach (maybe just need to adjust the time interval, POSTSTIM_STABILITY_EPOCH)

[...]

PR #511, opened 2021-04-12
GH #508 - Remove np.nan values from np.diff array

ru57y34nn:
# Overview:
np.nan values were not being dropped from the end of the np.diff array
before calling np.flatnonzero to remove non-zero values in get_stim_epoch.
This was causing get_stim_epoch to incorrectly identify the end of the
stimulus epoch as the same as the end of the sweep epoch, which was also
causing the end of the experiment epoch to be incorrect as well since it
relys on the stimulus epoch. By dropping the np.nan values from the end
of the difference array (di) in get_stim_epoch, the correct indices for
end of the stimulus epoch as well as for the experiment epoch are now being
returned. This should fix the issue with sweeps being incorrectly tagged
with 'Recording stopped before completing the experiment epoch'.

# Addresses:
Addresses issue https://github.com/AllenInstitute/ipfx/issues/508

# Type of Fix:

- [ ] Documentation Change

# Solution:
After getting the differences from adjacent data points in the current trace (i)
with np.diff(), the resulting np.nan values are dropped from the end of the array
with ~np.isnan() before the non-zero points are removed with np.flatnonzero().
There are np.nan points at the end of the np.diff() array because there are np.nan
values at the end of the current trace (i), and the np.diff() function returns np.nan
for the difference between two np.nan points, and then these np.nan points get 
picked up along with the up/down indices of the stimulus and the index of the 
last np.nan value will be mistaken as the ending index of the stimulus epoch. 

# Changes:
Drop np.nan from di array in get_stim_epoch()
Updated test_get_stim_epoch() in test_epochs.py to test array containing np.nan
Updated Unreleased section of CHANGELOG.md

# Validation:
This looks like it was just missed in testing. An array containing np.nan
values was not tested in the original unit test. The solution was a pretty 
simple so the added unit test should be sufficient. 
*np.nan values can only occur at the end of a sweep so only the  ending 
index of the stimulus epoch was affected. 

### Screenshots:
See GH issue #508 for screenshots
### Unit Tests:
See changes to test_get_stim_epoch() in test_epochs.py
### Configuration details:

# Checklist
- [x] My code follows
- [ ] My code is well-documented, and the docstrings conform to
      appropriate
- [x] The header on my commit includes the issue number
- [x] I have updated the CHANGELOG.md with the description of changes understandable by end users

# Notes:
Use this section to add anything you think worth mentioning to the
reader of the issue

### DIFF
--- a/ipfx/epochs.py
+++ b/ipfx/epochs.py
@@ -114,7 +114,8 @@ def get_stim_epoch(i, test_pulse=True):
     if test_pulse:
         di_idx = di_idx[2:]     # drop the first up/down (test pulse) if present
 
-    if len(di_idx) == 0:    # if no stimulus is found
+    # if no stimulus is found, len(di_idx=0), or end of stimulus is not found, len(di_idx=1)
+    if len(di_idx) < 2:    
         return None
 
     start_idx = di_idx[0] + 1   # shift by one to compensate for diff()
--- a/ipfx/epochs.py
+++ b/ipfx/epochs.py
@@ -108,6 +108,7 @@ def get_stim_epoch(i, test_pulse=True):
     """
 
     di = np.diff(i)
+    di = di[~np.isnan(di)] # != NaN
     di_idx = np.flatnonzero(di)   # != 0
 
     if test_pulse:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,12 @@
 All notable changes to this project will be documented in this file.
 
 ## Unreleased
-
+This pull request adds a line of code to the get_stim_epoch function in epochs.py to 
+remove np.nan values from the array of differences between data points before removing 
+all non-zero values. This change allows get_stim_epoch to find the correct ending index 
+of the stimlus epoch and the correct ending index of the experiment epoch as well since
+it relys on the stimulus epoch, which should fix the issue of sweeps being incorrectly 
+tagged with "Recording stopped before completing the experiment epoch".
 ### Added
 
 ### Changed
--- a/ipfx/qc_feature_extractor.py
+++ b/ipfx/qc_feature_extractor.py
@@ -307,7 +307,8 @@ def check_sweep_integrity(sweep, is_ramp):
             if sweep.epochs["recording"][1] < sweep.epochs["experiment"][1]:
                 tags.append("Recording stopped before completing the experiment epoch")
 
-    if np.isnan(sweep.i).any():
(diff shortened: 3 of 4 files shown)


tmchartrand:
@ru57y34nn a better fix for that last issue would be to change the stimulus epoch finding code to correctly recognize the epoch is incomplete, rather than returning out of order indices. https://github.com/AllenInstitute/ipfx/blob/0fc0a7452c831c3bb22e4d7dc3c767d17f9fade5/ipfx/epochs.py#L116
could probably just be changed to look for at least two elements (onset and offset).

ru57y34nn:
@tmchartrand Thanks for the suggestion, that should be a pretty easy change to make, I'll take a look at it this afternoon. Tim Jarsky pointed out that the epoch information is in the lab notebook of the pxp file and I would like to see if this also makes it into the nwb file. I am thinking that we should try to get the epoch information from the nwb file when possible as the current methods for finding the stimulus epoch will not work very will with more complicated stimulus waves.

ipfx/epochs.py
```diff
@@ -108,12 +108,14 @@ def get_stim_epoch(i, test_pulse=True):
     """
 
     di = np.diff(i)
+    di = di[~np.isnan(di)] # != NaN
```
sgratiy, line 111: This works as a quick fix to unblock you, but long term we should get epochs from the sweep table in the nwb.
Couple of issues:
1. It makes more sense to remove np.nan from the `i` array rather than from the processed `di`
2. If np.nan were to appear in the middle of an array like so:
[0, 0, 1, 1, 0, np.nan, 0, 2, 2, 2, 0, 0, np.nan, np.nan]
then you fix would also drop it. Not sure if that is a feasible case, but if it is, then it is better to just drop the trailing np.nan only and let the code fail down stream to catch this case. Something like this should work: 
```python
import ipfx.time_series_utils as tsu
end_idx = tsu.flatnotnan(i)[-1]
i=i[0:end_idx]


```

ru57y34nn: Thanks for the reply @sgratiy, and I agree with you about issue 1 that it makes more sense to remove np.nan from i rather than di. 

I also see your point for issue 2, although I don't believe there is anyway that we can have np.nan points in the middle of an array with data points before and after; we should only ever have trailing np.nan points, but I think it would be fine to use the method that you have shown here which should protect against this in case it is ever able to happen in the future.

tmchartrand:
Closing as #519 was merged instead.

PR #520, opened 2021-05-19
Address nan issue by selecting recording epoch as default

tmchartrand:
# Overview:
One of two options to address issue #508 by selecting the recording epoch by default as sweeps are created.

# Type of Fix:

- [ ] Documentation Change

# Solution:
The recording epoch already properly detects the valid portion of sweeps, but isn't always used in the early stages of QC or when detecting the stim epoch. By making sure it's always used, the full sweep info from the NWB is retained, but all code runs properly by dealing with the valid portion only.

# Changes:
- selects recording rather than sweep epoch by default, and removes redundant calls to select recording epoch
- restructures epoch detection to use recording epoch for stim epoch detection also
- restores previous behavior for missing data represented as zeros, replacing with nan rather than truncating
- fixes tests for compatibility with changes
- improved plot_data_set for quick visualization of new NWB2 to assess epochs

# Validation:
Checked that tests pass and that epochs are found correctly without error in a new MIES NWB2 file using the updated plot_data_set
```
from ipfx.bin.plot_ephys_nwb import plot_data_set
from ipfx.dataset.create import create_ephys_data_set
path = '/allen/programs/celltypes/production/humancelltypes/prod148/Ephys_Roi_Result_1077533543/H21.26.402.12.58.01.05.nwb'
data_set = create_ephys_data_set(path)
plot_data_set(data_set, clamp_mode='CurrentClamp', show_amps=True, qc_sweeps=True)
```

# Checklist
- [x] My code follows
- [x] My code is well-documented, and the docstrings conform to
      appropriate
- [ ] The header on my commit includes the issue number
- [ ] I have updated the CHANGELOG.md with the description of changes understandable by end users

### DIFF
--- a/ipfx/sweep.py
+++ b/ipfx/sweep.py
@@ -14,7 +14,7 @@ class Sweep(object):
         else:
             self.epochs = {}
 
-        self.selected_epoch_name = "sweep"
+        self.selected_epoch_name = "recording"
 
         if self.clamp_mode == "CurrentClamp":
             self.response = self._v
--- a/ipfx/sweep.py
+++ b/ipfx/sweep.py
@@ -14,16 +14,15 @@ class Sweep(object):
         else:
             self.epochs = {}
 
-        self.selected_epoch_name = "recording"
-
         if self.clamp_mode == "CurrentClamp":
-            self.response = self._v
-            self.stimulus = self._i
+            self._response = self._v
+            self._stimulus = self._i
         else:
-            self.response = self._i
-            self.stimulus = self._v
+            self._response = self._i
+            self._stimulus = self._v
 
         self.detect_epochs()
+        self.selected_epoch_name = "recording"
 
     @property
     def t(self):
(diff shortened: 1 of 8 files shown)


ipfx/sweep.py
```diff
@@ -54,22 +53,23 @@ def detect_epochs(self):
...
-            "experiment": ep.get_experiment_epoch(self._i, self.sampling_rate,test_pulse),
-            "stim": ep.get_stim_epoch(self.stimulus, test_pulse),
-        }
-
-        for epoch_name, epoch_detector in epoch_detectors.items():
-            if epoch_name not in self.epochs:
-                self.epochs[epoch_name] = epoch_detector
+        if "sweep" not in self.epochs:
+            self.epochs["sweep"] = ep.get_sweep_epoch(self._i)
+        self.select_epoch("sweep")
```
gouwens, line 64: Is this line necessary since the epoch gets changed to "recording" three lines down?

gouwens:
Overall it looks good to me - I just made one comment on a line that seems maybe left over. Thanks, @tmchartrand !

ipfx/sweep.py
tmchartrand: yeah you're right, was a holdover from a previous version before i simplified the sequence here, i'll remove.

MatthewAitken (approved):

Issue #537, opened 2021-11-22
IVSCC pipeline sweep extraction crashing with TypeError at get_nwb_version()


ru57y34nn:
**Describe the bug**
All IVSCC experiments have recently started failing sweep extraction with the same error. The following is the last few lines of the traceback from a EPHYS_NWB_STIMULUS_SUMMARY_V3_QUEUE log file:

File "/allen/aibs/technology/conda/production/fx/lib/python3.6/site-packages/ipfx/dataset/create.py", line 72, in get_nwb_version
    re.match("^2", nwb_version) or
  File "/allen/aibs/technology/conda/production/fx/lib/python3.6/re.py", line 172, in match
    return _compile(pattern, flags).match(string)
TypeError: cannot use a string pattern on a bytes-like object

The type error is being caused by the nwb_version attribute that get_nwb_version() gets from the nwb file to determine the version. The nwb_version attribute is expected to be a string but in recent nwb files the nwb_version attribute is a byte string which is causing re.match() to throw a TypeError as it is comparing a string with a byte object.

This issue seems to have arisen since the IVSCC pipeline upgraded from Igor Pro 8 to Igor Pro 9. This was an unexpected consequence of the upgrade.

**To Reproduce**
See EPHYS_NWB_STIMULUS_SUMMARY_V3_QUEUE log files for all recent IVSCC experiments since around 11/10/2021.

**Expected behavior**
the function get_nwb_version() should first check if nwb_version is an instance of the byte class and decode it to a string if necessary before using re.match() for comparison.

**Actual Behavior**
What actually happened. If the bug produced an error message or incorrect values, please include them here!

Full traceback from recent EPHYS_NWB_STIMULUS_SUMMARY_V3_QUEUE log file.
Traceback (most recent call last):
  File "/allen/aibs/technology/conda/production/fx/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/allen/aibs/technology/conda/production/fx/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/allen/aibs/technology/conda/production/fx/lib/python3.6/site-packages/ipfx/bin/run_sweep_extraction.py", line 96, in <module>
    main()
  File "/allen/aibs/technology/conda/production/fx/lib/python3.6/site-packages/ipfx/bin/run_sweep_extraction.py", line 89, in main
    module.args.get("stimulus_ontology_file", None)
[...]

PR #538, opened 2021-11-22
GH #537 - Decode nwb_version if byte object

ru57y34nn:
Add get_string_value() function to check if nwb_version attribute is
an instance of byte class and decode to string if necessary. This will
prevent a TypeError crash when get_nwb_version() calls re.match() with
nwb_version and is expecting a string and not a utf-8 encoded byte object.

# Overview:
Since moving from Igor Pro 8 to Igor Pro 9 all recent nwb version 2 files are storing nwb_version attribute as utf-8 encoded byte object rather than a string object. This was an unintended consequence of the upgrade and we are not yet sure why this changed but the fix is very simple. this pull request adds a new function called get_string_value() that is now used by get_nwb_versio() to first check if the nwb_version attribute that it gets from the nwb file is an instance of the byte class, specifically a utf-8 encode byte string, and converts it to a string if necessary before using it to check the version of the nwb file.

# Addresses:
Addresses issue #537 https://github.com/AllenInstitute/ipfx/issues/537

# Type of Fix:

- [x ] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation Change

# Solution:
This solution implements a check of the nwb_version attribute that comes from the nwb file and checks to see if it is an instance of the byte class and decodes it to a string if it is. This solution prevents the TypeError that is occurring with all pipeline nwb2 files before sweep extraction can be performed.

# Changes:
Very simple fix that just implements a simple function, get_string_value(), to be used by get_nwb_version to first check if the nwb_version attribute is a byte object and decode it to a string if it is.

# Validation:
This solution does not fail any current tests and I have been using it offline to process nwb files and extract features.

### Configuration details:

# Checklist
- [x ] My code follows
- [ x] My code is unit tested and does not decrease test coverage
- [ x] I have performed a self review of my own code
- [ x] My code is well-documented, and the docstrings conform to
- [ x] I have updated the documentation of the repository where
      appropriate
- [x ] The header on my commit includes the issue number
- [ x] My code passes all tests
- [ x] I have updated the CHANGELOG.md with the description of changes understandable by end users

# Notes:
This is a very simple fix and we are hoping to get this merged as quickly as possible as all IVSCC pipeline data is currently crashing prior to sweep extraction because of the issue that this pull request is addressing.


sgratiy:
@ru57y34nn I have made some changes. Please take a look at my PR against your branch.

sgratiy (approved):

PR #553, opened 2023-01-11
detect pauses update 

mzabolocki:
spkf module does not contain detect_pauses. Pause detection issue.


mzabolocki:
Realised spike_features.py did not contain detect_pauses module. Rather, it is in spike_train_features.py.

Issue #558, opened 2023-11-15
pip install ipfx fails on macOS with M1 arm64 cpu (apple silicon)


cudmore:
Hi all, I am trying to install ipfx on an m1 macOS (apple silicon) and the install fails.

My goal is to install ipfx so I can interoperate with ipfx from my desktop GUI called [SanPy](https://cudmore.github.io/SanPy/). Just had a poster at SFN and some Allen folks stopped by. Would be great if I could start using the 100+ features extracted by ipfx for current clamp recordings!

**Describe the bug**
`pip install ipfx` fails on macOS with M1 arm64 cpu (apple silicon)

**To Reproduce**

```
conda create -y -n tmp-env python=3.9
conda activate tmp-env
pip install ipfx
```

**Expected behavior**
Install should succeed

**Actual Behavior**
See pip install generated error pasted below

**Environment (please complete the following information):**
 - OS & version: macOS Ventura 13.3.1
 - Python version 3.9 or 3.11
 - AllenSDK version ???

**Here is the output of `pip install ipfx`**

```
(tmp-39) cudmore SanPy:pip install ipfx
Collecting ipfx
  Using cached IPFX-1.0.8-py2.py3-none-any.whl.metadata (2.9 kB)
Collecting argschema<2.0.0 (from ipfx)
  Using cached argschema-1.17.5-py2.py3-none-any.whl (18 kB)
Collecting allensdk (from ipfx)
  Using cached allensdk-2.16.1-py3-none-any.whl.metadata (1.9 kB)
Collecting dictdiffer (from ipfx)
  Using cached dictdiffer-0.9.0-py2.py3-none-any.whl (16 kB)
Collecting h5py==2.10.0 (from ipfx)
  Using cached h5py-2.10.0.tar.gz (301 kB)
  Preparing metadata (setup.py) ... done
Collecting marshmallow==3.0.0rc6 (from ipfx)
  Using cached marshmallow-3.0.0rc6-py2.py3-none-any.whl (42 kB)
Collecting matplotlib>=1.4.3 (from ipfx)
  Using cached matplotlib-3.8.1-cp39-cp39-macosx_10_12_x86_64.whl.metadata (5.8 kB)
Collecting methodtools (from ipfx)
  Using cached methodtools-0.4.7.tar.gz (3.6 kB)
  Preparing metadata (setup.py) ... done
Collecting numpy<1.19.0,>=1.15.4 (from ipfx)
[...]

Issue #557, opened 2023-07-17
can't install ipfx


sayakaminegishi:
**Describe the bug**
I tried to install ipfx by creating a conda environment and using pip install, but the python code I am trying to run still says "no module named 'ipfx'" and I cannot run code in my python program, such as "from ipfx.feature_extractor import (SpikeFeatureExtractor,SpikeTrainFeatureExtractor)"

It would be truly appreciated if you could help me fix this issue.
Thank you so much.

**To Reproduce**
(base) sayakaminegishi@Sayakas-MacBook-Pro ~ %conda create -n myenv python=3.6
(base) sayakaminegishi@Sayakas-MacBook-Pro ~ % conda activate myenv
(myenv) sayakaminegishi@Sayakas-MacBook-Pro ~ % pip install ipfx

**Expected behavior**
What you expected to happen.
be able to import ipfx in my python program.

**Actual Behavior**
unable to use/import ipfx. python console output gave "ModuleNotFoundError: No module named 'ipfx'"

**Environment (please complete the following information):**
 - OS & version: mac os
 - Python version 3.6?
 - AllenSDK version ?

**Additional context**

**Do you want to work on this issue?**
Are you willing and able to fix this bug? If so, let us know here (and see [the guide](https://github.com/AllenInstitute/AllenSDK/blob/master/CONTRIBUTING.md)). Thank you!

VeritatemAmo:
I also have problems installing on a M1 MacBook Pro using pip, with python version 3.9.18 and setuptools version 69.0.2, and cloning the master branch and run setup.py also no use. The error message for pip: 

  Preparing metadata (pyproject.toml) ... error
  error: subprocess-exited-with-error
  
  × Preparing metadata (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [54 lines of output]
      Running from numpy source directory.
      <string>:461: UserWarning: Unrecognized setuptools command, proceeding with generating Cython sources and expanding templates
      /private/var/folders/m2/syn15js158z7z32g0s_9dm1r0000gn/T/pip-install-z_806v96/numpy_44d7bb6586e04f3fb237f9725728b700/tools/cythonize.py:75: DeprecationWarning: distutils Version classes are deprecated. Use packaging.version instead.
        required_version = LooseVersion('0.29.14')
[...]

Issue #550, opened 2022-10-27
Can't pip install (dependency conflicts with AllenSDK?)


tmchartrand:
I'm not sure if this is universal, but currently a direct pip install of IPFX fails in a fresh python 3.8 conda environment. As far as I can tell, some dependency conflicts are forcing it back to an older AllenSDK version, which then requires an older version of tables (3.5.1) which has some install issues.

I'm guessing there are some requirements on both sides that could be loosened, but ultimately the best solution is probably to eliminate dependency on AllenSDK by breaking out the core modules that are actually used by other packages like IPFX. I've filed an issue there for that: https://github.com/AllenInstitute/AllenSDK/issues/2585.

guoyanzhao:
I was trying to install IPFX using recommended "pip install ipfx" but it failed. Lots of dependency issues. Could you please export your environment configuration so that other people can use that to create a conda enviroment that works?

sheriferson:
The PR to update IPFX Python version support and update most dependencies has been merged. This should resolve the installation issues you're facing if you install from the master branch. That said, the version hasn't been published to PyPI yet.

sheriferson:
IPFX 2.0.0 is now on PyPI and supports Python 3.9 to 3.11.

I think this should resolve the issues you are seeing @tmchartrand and @guoyanzhao. I will close this issue, but please reopen if there are still issues.

PR #559, opened 2023-11-17
Python 3.9/3.11 support, dropping Python 3.6 support, and moving to GitHub Actions

sheriferson:
# Overview:

This PR does the following:

- Drops support for Python 3.6
- Updates dependencies and code to run on Python 3.9/3.11
- Moves IPFX tests to GitHub Actions (and runs onprem tests using self-hosted runner)

# Type of Fix:

- [x] Breaking change (fix or feature that would cause existing
- [x] Documentation Change

# Unit Tests:

See here for the light and onprem test results from my fork:
https://github.com/sheriferson/ipfx/actions/runs/6898536678

# Checklist

- [x] My code follows
- [ ] My code is well-documented, and the docstrings conform to
      appropriate
- [ ] The header on my commit includes the issue number
- [x] I have updated the CHANGELOG.md with the description of changes understandable by end users

# Notes

I believe this addresses #558, #557, and #550. With the move away from Python 3.6 and the unpinning of most dependencies, `pip install -e .` succeeds on my M* Mac.

# Action items

- [ ] Deprecate CircleCI as well (?)
- [ ] Nightly workflow not detected by GitHub Actions yet. Maybe after merging into the main branch.
- [ ] Disable/deprecate outgoing internal CI/CD


README.md
```diff
@@ -22,8 +22,6 @@ We welcome contributions! Please see our [contribution guide](https://github.com
 
 Deprecation Warning
 -------------------
-The 1.0.0 release of ipfx brings some new features, like NWB2 support, along with improvements to our documentation and testing. We will also drop support for
-- NWB1
-- Python 2
 
-Older versions of ipfx will continue to be available, but may receive only occasional bugfixes and patches.
+The 1.1.0 release of IPFX drops support for Python 3.6 which reached end of life and stopped receiving security updated on December 23, 2021.
```
njmei, line 26: If we're following Semantic Versioning rules, IPFX should be on 2.0.0 after these changes because we are dropping Python3.6 support.

docker/README.md
```diff
@@ -0,0 +1,21 @@
...
+This directory contains Dockerfiles for building images for running on-prem tests that require internal Allen Institute resources. On-prem tests use GitHub self-hosted runners that will run tests on docker images built from these Dockerfiles.
+
+Our light and on-prem tests are defined in [our workflow file](../.github/workflows/github-actions-ci.yml "Link to GitHub Actions workflow for light and on-prem tests").
+
+- See [here](https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners) for more information on self-hosted runners.
+- See [here](https://docs.github.com/en/actions/hosting-your-own-runners/adding-self-hosted-runners) for more information on adding self-hosted runners to a GitHub repository.
+
+## Building images
+
+If you are an Allen Institute developer, you will have instructions on how to access the machine running the IPFX self-hosted runner.
```
njmei, line 12: I think it would be good to flesh out the process of running things on a self-hosted runner since I'm not sure the science folks will have easy access to the info.

docker/py39/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM python:3.9
+
+RUN apt-get update \
+    && apt-get install -y \
+        hdf5-tools \
+        curl
```
njmei, line 6: Should add ` && rm -rf /var/lib/apt/lists/*` according to best practices:
https://docs.docker.com/develop/develop-images/instructions/

docker/py39/Dockerfile
```diff
@@ -0,0 +1,10 @@
...
+
+RUN curl https://packagecloud.io/install/repositories/github/git-lfs/script.deb.sh | bash
+
+RUN apt-get install git-lfs
```
njmei, line 10: These `RUN` commands should also be done together with `&&` to prevent caching the `git-lfs` curl downloaded files in an intermediate Docker image layer (which bloats the final image size)

ipfx/attach_metadata/sink/nwb2_sink.py
```diff
@@ -236,36 +235,3 @@ def serialize(self, targets: Optional[OneOrMany[Dict[str, Any]]] = None):
                 file_.write(self._data.getvalue())
 
         self._reload_nwbfile()
-
-
-def set_container_sources(
```
njmei, line 241: I'm going to invoke [Chesterton's Fence](https://fs.blog/chestertons-fence/) here and ask for an explanation of what this existing code does? Why is it safe to remove this now but not previously?

requirements.txt
```diff
@@ -1,17 +1,18 @@
...
+pandas
 pg8000
 pillow
-pyabf<2.3.0
-pynwb>=1.3.2,<2.0.0
-pyYAML<6.0.0
-scipy>=0.15.1
-simplejson>=3.10.0
+pyabf
+pynwb==2.2.0
```
njmei, line 13: Have you tried with `pynwb~=2.2`? For external users having this pinned too strictly may cause pain points?

tests/test_nwb_utils.py
```diff
@@ -0,0 +1,6 @@
+from ipfx.dataset.create import is_file_mies
+
+TEST_NWB_FILE_PATH = 'tests/data/nwb/Ctgf-T2A-dgCre;Ai14-495723.05.02.01.nwb'
+
+def test_is_file_mies():
+    assert is_file_mies(TEST_NWB_FILE_PATH) == True
```
njmei, line 6: Can you add a test to test for the other type? I think it was HBG?

requirements.txt
njmei, line 1: I think we should have `requirements.txt` be as flexible as possible in terms of dependency pinning. But we should have have our tests use a `pinned-requirements.txt` so that future maintainers don't get constantly hit with dependency update related test failures.

requirements.txt
sheriferson: Yes, later versions (2.3.0+) start to fail a couple of tests, so 2.2.0 is the maximum for now without further changes.

README.md
sheriferson: Sounds good.

docker/README.md
sheriferson: I'm working on internal documentation that will be shared with scientists when this work is handed over.

docker/py39/Dockerfile
sheriferson: Will do. Thanks for the link.

tests/test_nwb_utils.py
sheriferson: There is no specific test/check for HBG files. The usage is checking if the file is MIES, and if not it assumes it's HBG.
There is one nwb file in the repo that is not MIES, I could create a test that shows that the NWB data loads successfully when `HBGNWBData` class is used?

ipfx/attach_metadata/sink/nwb2_sink.py
sheriferson: I agree with the invocation.
But I've been looking through #361 and the commit making the changes (1aeb641) and I am yet to find a good source of documentation for why the method existed.
There are a couple of internal confluence pages that I'm trying to get access to in case they are enlightening.
If you have suggestions for where to look or who to check with, please let me know.

njmei (approved):

PR #580, opened 2025-08-20
Require pinned pynwb version again

### LINKED ISSUE #543: Discussion related to #542
@njmei 
The original idea was that ipfx and MIES use the same nwb schema version.
As there is currently no support for multiple schema versions in one NWB file. I also doubt that this can be ever done properly.

From the requirements.txt of pynwb 

https://github.com/AllenInstitute/ipfx/blob/8c9d00d3f8f87b2a4c3bce29df4fe8140cdf5c65/requirements.txt#L13

and dandi-cli

https://github.com/dandi/dandi-cli/blob/39dc4f55caec0c92cdfed24db1bc604f988fc0e7/setup.cfg

one can work with older pynwb versions as well.

MIES supports NWB schema version 2.2.0, see https://github.com/AllenInstitute/IPNWB/blob/main/index.rst.

So in essence I think using the corresponding old pynwb version in ipfx, something like 1.2.1, would reduce the fragility introduced with #542.

t-b:
- [x] #582
- [x] Add integration tests with newer NWBv2 files, see also https://aiephys.slack.com/archives/C097D6GD6TV/p1755719172922809

Close #543

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -1,0 +1,29 @@
+# Testing
+
+The tests in the `tests` folder use pytest and can be run locally as:
+
+```
+SKIP_LIMS=true TEST_INHOUSE=false pytest
+```
+
+This does not assume that you have access to AllenInstitute internal resources like LIMS or network shares.
+
+## Strategy for adopting newer NWB schema versions
+
+The upstream source of NWB files for ipfx is [MIES](https://github.com/AllenInstitute/MIES) and
+[IPNWB](https://github.com/AllenInstitute/IPNWB). We are at the moment using fixed versions of nwb-schema for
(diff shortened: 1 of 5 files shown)


t-b:
Regarding the integration tests we'll do:

> Regarding the planned move to pynwb 3.2.1 for ipfx and the corresponding nwb-schema version 2.9.0  for MIES/IPNWB. Is there a test procedure with these new NWB files I should follow? The existing tests not requiring LIMS already pass locally.
I was involved in adopting NWBv2 in 2020, but we never did an schema version upgrade.
As an idea we could reexport the NWBv2 files used in test_run_feature_vector.py with the newer MIES/IPNWB version and run that test with these files as well.

Also document that procedure.

gouwens (approved):

PR #582, opened 2025-08-27
 Get it working again for newer versions of dependencies

### LINKED ISSUE #561: Cannot build on Python 3.12
I am unable to install this piece of software on Python 3.12. 
I get two error messages. One concerns the fact that Python 3.12 does not contain distutils anymore, and the build process depends on distutils. The other error message is from a build failure for Numpy 1.23.5, which is the most recent version of Numpy that is compatible with pynwb 2.2 and AllenSDK. The build fails because of changed functionality of the build tools, I guess. 

Numpy 1.23.5 was released Nov 19, 2022.
 
It would be nice to extend compatibility to pynwb 2.3.2, which is the earliest version of pynwb which moves the threshold forward for numpy compatibility, and migrate AllenSDK to a more recent version of Numpy as discussed here. https://github.com/AllenInstitute/AllenSDK/issues/2747

### LINKED ISSUE #567: Update supported python versions in setup.py
**Describe the bug**
The readme says

> IPFX is now tested on Python 3.9 and higher.

but setup.py only advertises 3.9 and 3.11. One of them should be fixed.

**Do you want to work on this issue?**
Yes I can work on that.

t-b:
This is a collaborative work to decruft ipfx. The goal is to make it usable with newer python/numpy/pynwb versions.

The tests without LIMS pass here locally with python 3.9-3.14rc2, numpy 2.32 and pynwb 3.1.2.

```
PYNWB_NO_CACHE_DIR=1 SKIP_LIMS=true TEST_INHOUSE=false pytest tests
```

ToDos:
- [x] Get CI working again, test with python 3.9 to 3.14
- [x] Fix supported python versions in setup.py
- [x] Wait for https://github.com/NeurodataWithoutBorders/pynwb/pull/2118 and a new release
- [x] Fix on-premise tests
- [x] Extend on-premise tests to cover the same range of python versions

Close #561
Close #567


requirements.txt
gouwens, line 1: Did you want to pin this to a specific pynwb version (i.e. the most recent version out now)?

gouwens: Otherwise all looks good to me

t-b: Thanks for the review. 

> Did you want to pin this to a specific pynwb version (i.e. the most recent version out now)?

This will be done in https://github.com/AllenInstitute/ipfx/pull/580.

gouwens (approved):

